PR #13, opened 2023-05-04
Initial polls demo

patrick91:
The goal of this demo is to show how to use Apollo Client with next 13 in the following cases:

- [x] SSR (rendering on the server, reuse cache on the client and mutate on the client)
- [x] RSC (rendering on the server, no cache reuse, mutation + full refresh on the client)
- [ ] RSC+server actions (rendering on the server, mutation the server, full refresh on client), once they are announced and available

Here's a video, it's a pretty simple demo, but should be helpful to understand how to use Apollo in the various scenarios 😊

https://user-images.githubusercontent.com/667029/236234073-017e2557-2526-407e-b6c1-f7d72dc40290.mp4

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -2,6 +2,8 @@
 
 Just a quick explainer:
 
-* [./hack-the-supergraph-ssr](./hack-the-supergraph-ssr) is a simple app that is based on the [hack the supergraph demo app](https://github.com/apollographql/hack-the-supergraph) app. 
+* [./hack-the-supergraph-ssr](./hack-the-supergraph-ssr) is a simple app that is based on the [hack the supergraph demo app](https://github.com/apollographql/hack-the-supergraph) app.
   It only showcases SSR cases, but it's a good starting point for understanding how to use the library.
 * [./app-dir-experiments/](./app-dir-experiments) is an app we use to internally do experiments. It's not really an example app, but if you just want to play around, go ahead.
+* [./polls-demo](./polls-demo/) is a simple poll app that showcases both Server Components and Client Components in a single app.
+  It show the difference between using Apollo Client in Server Components and Client Components.


examples/polls-demo/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@apollo/experimental-nextjs-app-support": "workspace:^",
```
nicolad, line 12: @patrick91 can we have here the actual dependency, not the internal monorepo package?

phryneas:
@nicolad The idea is that you check out the repo and build it once - then you could also experiment with the package itself, make changes, see what happens etc.

nicolad:
@phryneas the same experience of quickly creating a project from a template as in the vercel repo would be welcomed

https://github.com/vercel/next.js/tree/canary/examples/api-routes-apollo-server-and-client

phryneas:
@nicolad Honestly, I don't like the idea. Apollo Client is one of maybe 20 libraries you will be using in the end. So you'd need 20 different templates, but you'd always have to choose one of them, because they all are a "full project".   
Do you take the one that sets Material UI up for you? Or the one for Apollo Client? Or the one with the translation or state management library of your choice?  
On top of that, a template such as that would have to be constantly be updated to stay up to date with the newest files generated by `create-next-app`, and there's a good chance it would fall behind on times and you'd get an outdated setup as a result.

Setting up this library isn't too complicated. 
You create one or two files (depending if you want RSC, SSR or both) and hook one component into your root layout. It takes 10 minutes, will always be up-to-date and can easily be combined with any other library.

I'd rather have people do this by hand until we get a magical technology that adds AC into your existing repository (and the chances for that are low, since something like that would be very complicated).

PR #23, opened 2023-05-12
Add codegen to polls' demo

patrick91:


examples/polls-demo/.gitignore
```diff
@@ -34,4 +34,4 @@ yarn-error.log*
 *.tsbuildinfo
 next-env.d.ts
 .vscode/
-__generated__/
\ No newline at end of file
+*.generated.ts
```
phryneas, line 37: I'd suggest we check in the generated files as well so people can just start using this before having to run the codegen.

patrick91: ok!

PR #32, opened 2023-06-07
bump to client beta, use hooks w/o _experimental

phryneas:
Already opening this PR, but we should get https://github.com/apollographql/apollo-client/issues/10938 sorted out before we merge this.


jerelmiller (approved):
Approving so you can merge when the next beta release is available.

Issue #31, opened 2023-06-07
How does `SchemaLink` fit in here?


luchillo17:
Grabbing the example for Next.js [with-apollo-neo4j-graphql][1] before the `App` router became stable it used to check `isServer` (`typeof window === 'undefined'`) and based on that create a `SchemaLink`, which was supposed to use the same GraphQL query as a client instance, but be more efficient (I think it skips the HTTP request and queries the Schema directly, which in this example would be a Neo4jGraphQLSchema, not sure so don't quote me) but also limited to server-side components like SRC or SSR without the "use client" pragma (or whatever this metadata is called).

I'm wondering how this fits in the library since the README.md doesn't mention `SchemaLink`.

[1]: https://github.com/vercel/next.js/blob/canary/examples/with-apollo-neo4j-graphql/apollo/client.ts#L10-L11

jerelmiller:
Hey @luchillo17 👋 

First of all, take some of those examples with a grain of salt. I'm not sure how up-to-date those are, nor do I think those examples make use of this experimental library for proper RSC support. Some of them might still be fine, just know we haven't vetted those examples in quite some time.

Beyond that, I'm curious what you're looking for exactly when you say:

> I'm wondering how this fits in the library

There is no reason you can't or shouldn't use `SchemaLink`, its more that it just wasn't mentioned in this README. We instead wanted to focus on how this library bridges the gap with RSC and the new React paradigms that Next.js utilizes in Next.js 13.

luchillo17:
Hi @jerelmiller, by `wondering` I would like to see a full example project with the SchemaLink & the app router in use, if it can effectively skip the API request then I would consider a good pattern to encourage.

phryneas:
You do bring up a great point here - I think it might make sense to have this integrated in one of our examples, as it also makes it possible to use the `app` directory for GraphQL api hosting. 
See #36.

luchillo17:
I see we have a new `ApolloNextAppProvider`, still have yet to see `SchemaLink` mentioned anywhere in the README.

Also, the `neo4j-graphql` example I mentioned in the description was deleted, shame...

PR #36, opened 2023-06-09
(do not merge) schemalink in app-dir-experiments

phryneas:
@patrick91 This came up in #31 and could be an elegant solution to have the GraphQL api in a Route Handler file, while using it from the server (and also having it accessible during build).
I don't really want to add it to the `app-dir-experiments` example, but it might be worth adding that to the `polls` demo - what do you think?

### DIFF
--- a/examples/app-dir-experiments/app/ssr/ApolloWrapper.tsx
+++ b/examples/app-dir-experiments/app/ssr/ApolloWrapper.tsx
@@ -6,21 +6,18 @@ import {
   HttpLink,
   SuspenseCache,
 } from "@apollo/client";
+import { SchemaLink } from "@apollo/client/link/schema";
 import {
   ApolloNextAppProvider,
   NextSSRInMemoryCache,
   SSRMultipartLink,
 } from "@apollo/experimental-nextjs-app-support/ssr";
 import { setVerbosity } from "ts-invariant";
+import { schema } from "../api/graphql/route";
 
 setVerbosity("debug");
 
 function makeClient() {
-  const httpLink = new HttpLink({
-    uri: "http://localhost:3000/api/graphql",
-    fetchOptions: { cache: "no-store" },
-  });
-
   return new ApolloClient({
     cache: new NextSSRInMemoryCache(),
     link:
@@ -29,9 +26,14 @@ function makeClient() {
             new SSRMultipartLink({
               stripDefer: true,
             }),
-            httpLink,
+            new SchemaLink({
+              schema,
+            }),
           ])
-        : httpLink,
+        : new HttpLink({
+            uri: "http://localhost:3000/api/graphql",
+            fetchOptions: { cache: "no-store" },
+          }),
   });
 }
 
--- a/examples/app-dir-experiments/app/ApolloClient.ts
+++ b/examples/app-dir-experiments/app/ApolloClient.ts
@@ -1,14 +1,11 @@
-import { ApolloClient, HttpLink, InMemoryCache } from "@apollo/client";
+import { ApolloClient, InMemoryCache } from "@apollo/client";
+import { SchemaLink } from "@apollo/client/link/schema";
 import { registerApolloClient } from "@apollo/experimental-nextjs-app-support/rsc";
+import { schema } from "./api/graphql/route";
 
 export const { getClient } = registerApolloClient(() => {
   return new ApolloClient({
     cache: new InMemoryCache(),
-    link: new HttpLink({
-      uri: "http://localhost:3000/api/graphql",
-      // you can disable result caching here if you want to
-      // (this does not work if you are rendering your page with `export const dynamic = "force-static"`)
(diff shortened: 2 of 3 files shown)


patrick91:
this is quite interesting, I think I can change the demo to use this, maybe next week if I have time 😊

luchillo17:
If I remember correctly there are 3 environments where data may be fetched (correct me if I'm missing anything):

1. RSC (React Server Components)
2. SSR (Server Side Rendering)
3. CSR (Client-side rendering): This one is the outlier, can't bypass the HTTP request, and needs a server up & running to fetch the data.

Afaik `SchemaLink` would work for queries on RSC and SSR since it would bypass the HTTP request and query the resolver/db directly, however here's where I lack knowledge on these, how would a [mutation](https://www.apollographql.com/docs/react/data/mutations/) or a [subscription](https://www.apollographql.com/docs/react/data/subscriptions) work on those?

Let's say a Subscription is not a common use case, the one I'm worried about is the mutation, have yet to see an example of a mutation outside CSR, and an app without mutations is basically a static website...

luchillo17:
Went looking a bit more and SSR is basically the same as CSR aside from the cache rehydration after the first render, which is why we have a condition to check if running on server we use `SchemaLink`, on the client we use `HttpLink`, so a mutation would work like a normal client app right?

So the actual outlier here would be RSC, dunno what a mutation would look like in this case...

examples/app-dir-experiments/app/ssr/ApolloWrapper.tsx
```diff
@@ -6,21 +6,18 @@ import {
...
   SuspenseCache,
 } from "@apollo/client";
+import { SchemaLink } from "@apollo/client/link/schema";
 import {
   ApolloNextAppProvider,
   NextSSRInMemoryCache,
   SSRMultipartLink,
 } from "@apollo/experimental-nextjs-app-support/ssr";
 import { setVerbosity } from "ts-invariant";
+import { schema } from "../api/graphql/route";
```
luchillo17, line 16: I'm wondering about how importing `schema` here could expose credentials, `schema` with resolvers would need access to the DB driver, which in turn would need some credentials, we have a check for `isServer` so `SchemaLink` is only used in the server side, but the `schema` is imported for both, meaning the bundle might end up with the driver setup in the client, and depending on how it's bundled it would have the credentials embedded somewhere, do you think I'm making a correct assumption here?

phryneas: From my experiences with the Nextjs bundler, I would assume that the `typeof window !== "undefined"` check would remove `SchemaLink` as well as `schema` from the client bundle shipped to the browser, but I'd be grateful if you could validate that further :)

luchillo17:
Hmm I don't think mutation with RSC works yet, when I try to execute a mutation inside a server action I get an error saying I should mark it as `"use server"`, however I am already doing it, it breaks specifically when I add the mutation code, not sure what the issue is, guess it's marked as experimental for a reason:
image

luchillo17:
My guess at this point is that `ServerActions` is experimental, RSC currently only works with data fetching but interactivity (like `onClick` or form submission) right now is client-side only, so currently for an app with mutations we would have to use the SSR version of this library, or find a fix for the error above, which I wouldn't recommend until `ServerActions` gets released (out of experimental phase), the docs themselves mention the use cases for RSC and the rest:
https://nextjs.org/docs/getting-started/react-essentials#when-to-use-server-and-client-components

examples/app-dir-experiments/app/ssr/ApolloWrapper.tsx
luchillo17: Can't validate, my setup uses Neo4j GraphQL schema, which is generated async, creating the wrapper with the schema makes it harder since the `makeClient` function would need to be async, and Next.js complains about functional components that can't be async, sorry.

examples/app-dir-experiments/app/ssr/ApolloWrapper.tsx
```diff
@@ -29,9 +26,14 @@ function makeClient() {
             new SSRMultipartLink({
               stripDefer: true,
             }),
-            httpLink,
+            new SchemaLink({
+              schema,
```
michalkvasnicak, line 30: what if we need to create a `context` here with `req` to determine user's authentication state in graphql resolvers?

examples/app-dir-experiments/app/ssr/ApolloWrapper.tsx
luchillo17: I am dumb, dunno why I didn't think about this before but to solve my issue of schema generator being async, I could just wrap `makeClient` into an async factory, declare `<ApolloWrapper>` component as async and inside the component, await the factory to get the `makeClient`, can't believe it took me this long to think about this solution...

PR #38, opened 2023-06-14
Adds support for `useBackgroundQuery` and `useReadQuery` in Streaming SSR

alessbell:
This PR creates a new class, `NextSSRApolloClient` in `package/src/ssr/NextSSRApolloClient.tsx`, which can be imported and used in **client components** where one would use `ApolloClient` outside of the app directory.

This extension of the `ApolloClient` class registers "background queries" (queries triggered by components rendering on the server) via the same `registerLateInitializingQueue` strategy already used to register results on the client. By keeping track of queries already run on the server, we can avoid duplicating these requests on the client when the same components are executed in the browser, since the result will be streamed in once it has completed.

### DIFF
--- a/examples/hack-the-supergraph-ssr/app/error.tsx
+++ b/examples/hack-the-supergraph-ssr/app/error.tsx
@@ -4,16 +4,14 @@ import React from "react";
 import { Box, Heading, Text, VStack } from "@chakra-ui/react";
 
 export const Error = ({
-  children,
-  code,
   error,
-}: React.PropsWithChildren<{
-  code?: string;
-  error: { toString(): string };
-}>) => (
+  reset,
+}: {
+  error: Error;
+  reset: () => void;
+}) => (
   <VStack spacing="12">
     <VStack textAlign="center">
-      <Heading size="4xl">{code ?? "Unknown Error"}</Heading>
       <Heading fontSize="3xl">Houston, something went wrong on our end</Heading>
       <Text>Please review the information below for more details.</Text>
     </VStack>
@@ -25,10 +23,9 @@ export const Error = ({
         borderRadius="8px"
         borderColor="brand.light"
       >
-        <Text color="brand.error">{error.toString()}</Text>
+        <Text color="brand.error">Error: {error.message}</Text>
       </Box>
     )}
-    {children}
   </VStack>
 );
 
--- a/examples/polls-demo/app/cc/loading.tsx
+++ b/examples/polls-demo/app/cc/loading.tsx
@@ -1,5 +1,0 @@
-import { PollSkeleton } from "@/components/poll";
-
-export default function Loading() {
-  return <PollSkeleton />;
-}
--- a/package/.eslintrc.js
+++ b/package/.eslintrc.js
@@ -20,5 +20,11 @@ module.exports = {
   rules: {
     "@typescript-eslint/no-explicit-any": "off",
     "react/prop-types": "off",
+    "@typescript-eslint/no-unused-vars": [
+      "warn",
+      {
+        varsIgnorePattern: "^_",
+      },
+    ],
   },
 };
--- a/package/README.md
+++ b/package/README.md
@@ -9,7 +9,7 @@
 > This cannot be addressed from our side, but would need API changes in Next.js or React itself.  
 > If you do not use suspense in your application, this will not be a problem to you.  
 
-> ❗️ This package depends on Apollo Client 3.8, which is currently an alpha release. 
+> ❗️ This package depends on Apollo Client 3.8, which is currently a beta release. 
 
 ## Detailed technical breakdown
 
@@ -31,10 +31,10 @@ This package provides the tools necessary to execute your GraphQL queries on the
 
 ## Installation
 
-This package has a peer dependency on the latest alpha of `@apollo/client`, so you can install both this package and that Apollo Client version via
+This package has a peer dependency on the latest beta of `@apollo/client`, so you can install both this package and that Apollo Client version via
 
 ```sh
-npm install @apollo/client@alpha @apollo/experimental-nextjs-app-support
+npm install @apollo/client@beta @apollo/experimental-nextjs-app-support
 ```
 
 ## Usage
@@ -172,6 +172,17 @@ You can import the following Apollo Client hooks from `"@apollo/experimental-nex
 - `useFragment`
 
 If you want to make the most of the streaming SSR features offered by React & the Next.js App Router, consider using the [`useSuspenseQuery`](https://www.apollographql.com/docs/react/api/react/hooks-experimental/#using-usesuspensequery_experimental) and [`useFragment`](https://www.apollographql.com/docs/react/api/react/hooks-experimental/#using-usefragment_experimental) hooks.
+
+### Resetting singletons between tests.
+This package uses some singleton instances on the Browser side - if you are writing tests, you must reset them between tests.
+
+For that, you can use the `resetNextSSRApolloSingletons` helper:
+
+```ts
+import { resetNextSSRApolloSingletons } from "@apollo/experimental-nextjs-app-support/ssr";
+
+afterEach(resetNextSSRApolloSingletons);
+```
 
 ## Handling Multipart responses in SSR
 
--- a/package/README.md
+++ b/package/README.md
@@ -82,7 +82,6 @@ First, create a new file `app/ApolloWrapper.js`:
 // ^ this file needs the "use client" pragma
 
 import {
-  ApolloClient,
   ApolloLink,
   HttpLink,
   SuspenseCache,
@@ -90,6 +89,7 @@ import {
 import {
   ApolloNextAppProvider,
   NextSSRInMemoryCache,
+  NextSSRApolloClient,
   SSRMultipartLink,
 } from "@apollo/experimental-nextjs-app-support/ssr";
 
@@ -103,7 +103,7 @@ function makeClient() {
     fetchOptions: { cache: "no-store" },
   });
 
-  return new ApolloClient({
+  return new NextSSRApolloClient({
     // use the `NextSSRInMemoryCache`, not the normal `InMemoryCache`
     cache: new NextSSRInMemoryCache(),
     link:
@@ -164,7 +164,12 @@ export default function RootLayout({
 > ☝️ This will work even if your layout is a React Server Component and will also allow the children of the layout to be React Server Components.  
 > It just makes sure that all Client Components will have access to the same Apollo Client instance, shared through the `ApolloNextAppProvider`.
 
-Now you can use the hooks `useQuery`, `useSuspenseQuery`, `useFragment`, and `useApolloClient` from `"@apollo/experimental-nextjs-app-support/ssr"` in your Client components like you are used to.
+You can import the following Apollo Client hooks from `"@apollo/experimental-nextjs-app-support/ssr"` in your client components (make sure you are not importing these hooks from `@apollo/client` as this package wraps and re-exports them to support streaming SSR):
+- `useQuery`
+- `useSuspenseQuery`
+- `useBackgroundQuery`
+- `useReadQuery`
+- `useFragment`
 
 If you want to make the most of the streaming SSR features offered by React & the Next.js App Router, consider using the [`useSuspenseQuery`](https://www.apollographql.com/docs/react/api/react/hooks-experimental/#using-usesuspensequery_experimental) and [`useFragment`](https://www.apollographql.com/docs/react/api/react/hooks-experimental/#using-usefragment_experimental) hooks.
 
--- a/package/src/ssr/ApolloRehydrateSymbols.tsx
+++ b/package/src/ssr/ApolloRehydrateSymbols.tsx
@@ -1,5 +1,9 @@
 import { SuperJSONResult } from "superjson/dist/types";
-import { RehydrationCache, ResultsCache } from "./types";
+import {
+  RehydrationCache,
+  ResultsCache,
+  BackgroundQueriesCache,
+} from "./types";
 import type { DataTransport } from "./dataTransport";
 
 declare global {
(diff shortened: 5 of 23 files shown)


examples/hack-the-supergraph-ssr/app/error.tsx
```diff
@@ -4,16 +4,14 @@ import React from "react";
...
-  children,
-  code,
   error,
-}: React.PropsWithChildren<{
-  code?: string;
-  error: { toString(): string };
-}>) => (
+  reset,
+}: {
+  error: Error;
```
alessbell, line 10: Updating our error pages to conform with the API as described in the latest docs: https://nextjs.org/docs/app/api-reference/file-conventions/error

alessbell: Interestingly, there's a lint error if I try to type `error` as `ApolloError` here:

CleanShot 2023-06-14 at 11 46 38

examples/polls-demo/app/cc/loading.tsx
```diff
@@ -1,5 +0,0 @@
-import { PollSkeleton } from "@/components/poll";
```
alessbell, line 1: I'll add this back - I temporarily removed it for testing with `useBackgroundQuery` where using a `Suspense` boundary directly was preferred

package/.eslintrc.js
```diff
@@ -20,5 +20,11 @@ module.exports = {
   rules: {
     "@typescript-eslint/no-explicit-any": "off",
     "react/prop-types": "off",
+    "@typescript-eslint/no-unused-vars": [
+      "warn",
+      {
+        varsIgnorePattern: "^_",
```
alessbell, line 26: Added this for array destructuring where the first item isn't needed

package/src/ssr/ApolloRehydrateSymbols.tsx
```diff
@@ -1,14 +1,23 @@
...
+  BackgroundQueriesCache,
+} from "./types";
 import type { DataTransport } from "./dataTransport";
 
 declare global {
   interface Window {
     [ApolloRehydrationCache]?: RehydrationCache;
     [ApolloResultCache]?: ResultsCache;
     [ApolloSSRDataTransport]?: DataTransport<SuperJSONResult>;
+    // TODO: types
```
alessbell, line 14: I think I already addressed this, will remove the comment

package/src/ssr/RehydrationContext.tsx
```diff
@@ -13,20 +14,18 @@ const ApolloRehydrationContext = React.createContext<
...
-    if (cache instanceof NextSSRInMemoryCache) {
-      cache.setRehydrationContext(rehydrationContext.current);
-    } else {
-      throw new Error(
-        "When using Next SSR, you must use the `NextSSRInMemoryCache`"
-      );
-    }
+    if (client instanceof NextSSRApolloClient) {
+      client.setRehydrationContext(rehydrationContext.current);
+    } // else error?
```
alessbell, line 25: Oops missed this comment, adding these errors shortly..

package/README.md
```diff
@@ -164,7 +164,12 @@ export default function RootLayout({
 > ☝️ This will work even if your layout is a React Server Component and will also allow the children of the layout to be React Server Components.  
 > It just makes sure that all Client Components will have access to the same Apollo Client instance, shared through the `ApolloNextAppProvider`.
 
-Now you can use the hooks `useQuery`, `useSuspenseQuery`, `useFragment`, and `useApolloClient` from `"@apollo/experimental-nextjs-app-support/ssr"` in your Client components like you are used to.
+You can import the following Apollo Client hooks from `"@apollo/experimental-nextjs-app-support/ssr"` in your client components (make sure you are not importing these hooks from `@apollo/client` as this package wraps and re-exports them to support streaming SSR):
```
alessbell, line 167: Just made these hooks a list to make this section a bit harder to miss.

I removed `useApolloClient` from this list since that seemed no longer relevant, but let me know if that should be added back @phryneas 👍

package/src/ssr/NextSSRApolloClient.tsx
```diff
@@ -0,0 +1,169 @@
...
+  > & { uninitialized?: boolean } = {
+    incomingBackgroundQueries: [],
+    uninitialized: true,
+  };
+
+  constructor(options: ApolloClientOptions<TCacheShape>) {
+    super({
+      ...options,
+      // TODO: the memoization in `documentTransform` can be removed once
+      // https://github.com/apollographql/apollo-client/pull/10968 is merged.
```
alessbell, line 37: This was just merged, I should be able to remove this once the next AC beta is released 🎉

alessbell: After upgrading to beta 3, the client-side fetch is no longer being deduped. Investigating...

package/src/ssr/NextSSRApolloClient.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const result = super.watchQuery(options);
+    return result;
+  }
+
+  setRehydrationContext(rehydrationContext: RehydrationContextValue) {
+    if (this.rehydrationContext.uninitialized) {
+      rehydrationContext.incomingBackgroundQueries.push(
+        ...this.rehydrationContext.incomingBackgroundQueries
+      );
+    }
```
phryneas, line 165: Maybe here we could even move all data from the old `incombineBackgroundQueries` to the new ones?

Something like

```suggestion
    if (rehydrationContext.incomingBackgroundQueries !== this.rehydrationContext.incomingBackgroundQueries)
      rehydrationContext.incomingBackgroundQueries.push(
        ...this.rehydrationContext.incomingBackgroundQueries.splice(0, Infinity)
      )
```

Then we could skip all the `uninitialized` handling altogether.

package/src/ssr/NextSSRApolloClient.tsx
```diff
@@ -0,0 +1,169 @@
...
+                this["queryManager"].fetchCancelFns.delete(cacheKey);
+
+              const [_, reject] = this.resolveFakeQueries.get(cacheKey) ?? [];
+
+              this["queryManager"].fetchCancelFns.set(
+                cacheKey,
+                (reason: unknown) => {
+                  cleanupCancelFn();
+                  if (reject) {
+                    this.resolveFakeQueries.delete(cacheKey);
```
phryneas, line 123: Maybe instead of calling this in multiple places, we could have a `promise.finally(...)` call with this directly after the definition of `promise`?

package/src/ssr/NextSSRApolloClient.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    .catch((err) => {
+                      observer.error(err);
+                    });
+                })
+              );
+
+              const cleanupCancelFn = () =>
+                this["queryManager"].fetchCancelFns.delete(cacheKey);
+
+              const [_, reject] = this.resolveFakeQueries.get(cacheKey) ?? [];
```
phryneas, line 116: This line should probably better go into the `fetchCancelFn` definition a few lines down - otherwise `reject` would never be undefined, and also execute if the query had resolved successfully.

package/src/ssr/NextSSRInMemoryCache.tsx
```diff
@@ -18,21 +16,6 @@ export class NextSSRInMemoryCache extends InMemoryCache {
...
-      if (Array.isArray(window[ApolloResultCache] || [])) {
-        registerLateInitializingQueue(ApolloResultCache, (data) =>
-          this.write(data)
-        );
-      } else {
-        throw new Error(
-          "On the client side, only one instance of `NextSSRInMemoryCache` can be created!"
-        );
-      }
-    }
```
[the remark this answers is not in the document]
phryneas: Huh, good question. If a real `watchQuery/fetchQuery` call latches onto our fake query before it finishes, it wouldn't be necessary - but if it doesn't, it would be necessary.
So maybe we could create the fake ongoing query, and then immediately call `QueryManager.fetchQuery` to make sure its' results gets written into the store? Or we just call `write` after resolving the query? (The latter might have a lower risk of race conditions 🤔 )

phryneas (changes requested):
Already some feedback. I'm curious why that deduplication change didn't do what we expected, though. Gonna look a little deeper 👀

phryneas:
@alessbell I found out what was going wrong.

We need a `getQuerymanager(this)` method instead of just accessing `this['QueryManager']`, as that one is `any`, not `QueryManager<TCacheShape>`. Add that and you'll immediately notice ;)

package/src/ssr/NextSSRApolloClient.tsx
alessbell: Fixed in [9a74dc6](https://github.com/apollographql/apollo-client-nextjs/pull/38/commits/9a74dc607f20aa1239cfdeee31f0e0eb157b5e43)

package/src/ssr/NextSSRInMemoryCache.tsx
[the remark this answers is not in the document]
alessbell: Sounds good :) Addressed in [9a74dc6](https://github.com/apollographql/apollo-client-nextjs/pull/38/commits/9a74dc607f20aa1239cfdeee31f0e0eb157b5e43).

package/src/ssr/NextSSRApolloClient.tsx
alessbell: Fixed in [c49d481](https://github.com/apollographql/apollo-client-nextjs/pull/38/commits/c49d481ff5c9bdd474c4b61ba65b610162ffe067).

package/src/ssr/NextSSRApolloClient.tsx
alessbell: Fixed in [c49d481](https://github.com/apollographql/apollo-client-nextjs/pull/38/commits/c49d481ff5c9bdd474c4b61ba65b610162ffe067)

package/src/ssr/NextSSRApolloClient.tsx
alessbell: Addressed in [fa82b94](https://github.com/apollographql/apollo-client-nextjs/pull/38/commits/fa82b94c0189747b9463628c1c57b8d74803d1af)

alessbell:
Thanks for the review, @phryneas! I believe I've gotten to all the feedback (including offline comment re: test helpers addressed in [037c8ba](https://github.com/apollographql/apollo-client-nextjs/pull/38/commits/037c8ba783feaf46f6e81d9fb3259142e9cdb67e)) and I've done some more manual testing. This is ready for a final look :)

examples/polls-demo/app/cc/loading.tsx
phryneas: Reminder to add this back :)

package/src/ssr/hooks.ts
```diff
@@ -21,6 +24,13 @@ export const useSuspenseQuery = wrap(orig_useSuspenseQuery, [
   "data",
   "networkStatus",
 ]);
+export const useReadQuery = wrap(orig_useReadQuery, ["data", "networkStatus"]);
+// @ts-ignore
+export const useBackgroundQuery: typeof orig_useBackgroundQuery = (...args) => {
+  useRehydrationContext();
+  // @ts-ignore
+  return orig_useBackgroundQuery(...args);
+};
```
phryneas, line 33: let's get rid of those `@ts-ignore` annotations so eslint shuts up :)
```suggestion
export const useBackgroundQuery: typeof orig_useBackgroundQuery = (...args: [any, any]) => {
  useRehydrationContext();
  return orig_useBackgroundQuery(...args);
};
```

package/src/ssr/NextSSRApolloClient.tsx
```diff
@@ -0,0 +1,169 @@
...
+                  promise
+                    .then((result) => {
+                      observer.next(result);
+                      observer.complete();
+                    })
+                    .catch((err) => {
+                      observer.error(err);
+                    }).finally(() => {
+                      this.resolveFakeQueries.delete(cacheKey);
+                    });
```
phryneas, line 105:
```suggestion
                    }).finally(() => {
                      this.resolveFakeQueries.delete(cacheKey);
                      cleanupCancelFn();
                    });
```

For comparison:
https://github.com/apollographql/apollo-client/blob/d64a9656ee745349bb4958a743065f2a91c2d0cc/src/core/QueryManager.ts#L1278

package/src/ssr/NextSSRApolloClient.tsx
```diff
@@ -0,0 +1,169 @@
...
+                })
+              );
+              const queryManager = getQueryManager<TCacheShape>(this);
```
phryneas, line 108:
```suggestion
```

This should already be in scope due to line 77.

phryneas (changes requested):
This looks very good! 
Only some minor little changes, then this should be good to go!

package/src/ssr/hooks.ts
alessbell: Resolved in [867d581](https://github.com/apollographql/apollo-client-nextjs/pull/38/commits/867d5812951b586c6c8cc23fba07e82e0dd129bf)

package/src/ssr/NextSSRApolloClient.tsx
alessbell: Ah yep, thanks for catching that! Resolved via [6462b2a](https://github.com/apollographql/apollo-client-nextjs/pull/38/commits/6462b2a8c2f1dfb24a66a34f77944cb1a9436546)

package/src/ssr/NextSSRApolloClient.tsx
alessbell: Resolved in [6462b2a](https://github.com/apollographql/apollo-client-nextjs/pull/38/commits/6462b2a8c2f1dfb24a66a34f77944cb1a9436546)

PR #42, opened 2023-06-26
call `cache.write` independently of `resolve` presence

phryneas:
I just noticed this one when setting up our integration tests.

https://github.com/apollographql/apollo-client/blob/pr/bundling-integrations/integration-tests/next/tests/playwright/apollo-client.test.ts#L10-L14


alessbell (approved):
Thanks for catching this! 🎉

Issue #48, opened 2023-06-28
Infinite loading with `useQuery`


MarkLauer:
Hello, I've been trying to use the package for SSR, but if I call `useQuery` in a client component, the request is not executed on the server side and there's infinite loading in the browser (transported data is always`{ data: undefined, loading: true }`). Also the browser doesn't execute the query too (however, I guess it is meant to be this way because the client considers the request already done on the server).

I've created a [demo](https://codesandbox.io/p/sandbox/eloquent-microservice-s2hwp4?file=%2Fapp%2Fpoll-cc.tsx) based on polls-demo example from this repository.

Also I want to say thank you to the contributors of this package, it is so convenient to have a first-class support of RSC/SSR for Apollo+Next.js combo.

quocanhbk:
I've encountered the same issue. It even keeps rendering infinitely and eventually freezes my browser. I set up the project based from this [article](https://www.apollographql.com/blog/announcement/frontend/using-apollo-client-with-next-js-13-releasing-an-official-library-to-support-the-app-router/)

Here is my home page code:
```ts
// app/page.tsx
"use client";

import { gql } from "@apollo/client";
import { useQuery } from "@apollo/experimental-nextjs-app-support/ssr";

const query = gql`
  query Acts {
    acts {
      id
      title
    }
  }
`;

export default async function Home() {
  const { data, loading, called } = useQuery(query);

  console.log({ data, loading, called });

  return (
    <div>
      <h1>Client Component</h1>
      <p>Loading: {loading ? "True" : "False"}</p>
      <p>Data:</p>
      <pre>{JSON.stringify(data, null, 2)}</pre>
    </div>
  );
}
```

Here is the repo to reproduce the issue: 
https://github.com/quocanhbk/nextjs-app-dir-test

Here is the result of infinite render:
<img width="481" alt="image" src="https://github.com/apollographql/apollo-client-nextjs/assets/64638082/a125315e-7a76-4273-a927-8ea6bb7cf7df">

phryneas:
So, here's the thing about `useQuery`: it won't start a request during SSR, because that would require it to suspend and substatially change the way that `useQuery` is expected to work. If you want to do the request during SSR, you'd need to use `useSuspenseQuery` or `useBackgroundQuery`.  
[...]

PR #49, opened 2023-06-29
fix `useQuery` bugs

phryneas:
Fixes a regression introduced by 0.3.0 - see #48

* prevent a network request on the server
* prevent transportation of non-network-request `watchQuery` calls


integration-test/fixture.ts
```diff
@@ -8,6 +37,7 @@ export const test = base.extend<{
     page.on("pageerror", (error) => {
       expect(error.stack || error).toBe("no error");
     });
+    page.route("**", (route) => route.continue());
```
phryneas, line 40: this prevents the playwright http cache to kick in in test development

integration-test/fixture.ts
```diff
@@ -1,5 +1,34 @@
...
+declare global {
+  namespace PlaywrightTest {
+    interface Matchers<R, T> {
+      toBeInitiallyLoading(expectedState: boolean): R;
+    }
+  }
+}
+
+expect.extend({
+  async toBeInitiallyLoading(
```
phryneas, line 12: This tests for the loading indicator **immediately** instead of waiting up to 5 seconds for it to appear or disappear, which will accidentally succeed in a lot of cases.

integration-test/fixture.ts
alessbell: Maybe add a comment to that effect (unless this is a well established Playwright convention)?

phryneas: good point :) And I have no idea ^^

phryneas, line 40:
```suggestion
    // this prevents the playwright http cache to kick in in test development
    page.route("**", (route) => route.continue());
```

integration-test/fixture.ts
```diff
@@ -1,5 +1,34 @@
...
+      return {
+        message: () => "passed",
+        pass: true,
+      };
+    } else {
+      return {
+        message: () => "failed",
+        pass: false,
+      };
+    }
```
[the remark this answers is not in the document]
phryneas: Good point :) Thanks!

PR #59, opened 2023-07-13
make sure to clean up `inFlightLinkObservables`

### LINKED ISSUE #56: Queries on first loaded page are never fetched again
**Problem**
Using the SSR ApolloWrapper as defined in the [README](https://github.com/apollographql/apollo-client-nextjs/blob/main/package/README.md), any page that has `useSuspenseQuery` will be filled with data (fetched on the server) right away. That data, however, is never refetched no matter what `fetchPolicy` you define. The pages are not cached (`export const dynamic = 'force-dynamic'`), but the query never seems to be run again.

Queries on other pages are fetched as expected, it's just the first page that is loaded (regardless of which page that is) will keep displaying the results of its initial SSR fetch.

**How to reproduce**
I have set up a [CodeSandbox](https://codesandbox.io/p/sandbox/interesting-dew-v7ccyr) to demonstrate this specific issue.

1. Open page one (hard refresh).
2. Note the total votes.
3. In a separate tab, [increment the vote count](https://apollo-next-13-polls.vercel.app/cc).
4. Visit page two, and go back to page one.
5. We expect the total count to first show the cached value, and then be updated with the latest value (`cache-and-network`). 
6. The value is never updated. :x:

We can repeat step 3 and 4 indefinitely, but page one will always show the initial total votes from step 2.

When page one **is not** the initial page we load, we get the expected behavior on page one:

1. Open page two (hard refresh).
2. Visit page one.
3. Note the total votes.
4. In a separate tab, [increment the vote count](https://apollo-next-13-polls.vercel.app/cc).
5. Visit page two, and go back to page one.
6. We expect the total count to first show the cached value, and then be updated with the latest value (`cache-and-network`).
7. The value is updated. :heavy_check_mark:

**Video of issue occurring in CodeSandbox**

https://github.com/apollographql/apollo-client-nextjs/assets/1277208/2cf749ef-f419-4ea8-8cc2-3e77c0bb93e4

**CodeSandbox**
https://codesandbox.io/p/sandbox/interesting-dew-v7ccyr

phryneas:
This should fix #56

### DIFF
--- a/package/src/ssr/NextSSRApolloClient.tsx
+++ b/package/src/ssr/NextSSRApolloClient.tsx
@@ -90,15 +90,36 @@ export class NextSSRApolloClient<
             );
 
             if (!byVariables.has(varJson)) {
+              let resolveFakeQuery: [
+                  (result: FetchResult) => void,
+                  (reason: any) => void
+                ],
+                observable: Observable<FetchResult>,
+                fetchCancelFn: (reason: unknown) => void;
+
+              const cleanup = () => {
+                if (
+                  queryManager["fetchCancelFns"].get(cacheKey) === fetchCancelFn
+                )
+                  queryManager["fetchCancelFns"].delete(cacheKey);
+
+                if (byVariables.get(varJson) === observable)
+                  byVariables.delete(varJson);
+
+                if (this.resolveFakeQueries.get(cacheKey) === resolveFakeQuery)
+                  this.resolveFakeQueries.delete(cacheKey);
+              };
+
               const promise = new Promise<FetchResult>((resolve, reject) => {
-                this.resolveFakeQueries.set(cacheKey, [resolve, reject]);
+                this.resolveFakeQueries.set(
+                  cacheKey,
+                  (resolveFakeQuery = [resolve, reject])
+                );
               });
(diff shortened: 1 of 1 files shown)


package/src/ssr/NextSSRApolloClient.tsx
```diff
@@ -90,15 +90,36 @@ export class NextSSRApolloClient<
...
+                fetchCancelFn: (reason: unknown) => void;
+
+              const cleanup = () => {
+                if (
+                  queryManager["fetchCancelFns"].get(cacheKey) === fetchCancelFn
+                )
+                  queryManager["fetchCancelFns"].delete(cacheKey);
+
+                if (byVariables.get(varJson) === observable)
+                  byVariables.delete(varJson);
```
phryneas, line 107: The `byVariables.delete(varJson);` is the relevant part of this PR - the rest is mostly cleanup.

phryneas:
The broken tests are due to #59 being unmerged.

Here is a working test from before rebasing: https://github.com/apollographql/apollo-client-nextjs/actions/runs/5543562205

jerelmiller (approved):

PR #68, opened 2023-07-19
Allow args on function passed to registerApolloClient

### LINKED ISSUE #67: How can I have clients for many different HTTP links, to implement draft previews?
The site I'm working on queries for data from a CMS which supports previews for drafts and upcoming releases. When the user requests a preview, the front end is given a token, and based on this we add some query parameters to the GQL endpoint URL.

In the past I've done this by having a map of preview tokens (or null) to Apollo clients, and then in my `getClient` function taking in as an argument the preview token (or null), returning an existing client from the map if there's a matching one, and otherwise making a new client, adding it to the map, and returning it.

I'm not sure how to do something equivalent with `registerApolloClient` provided by this package. My first thought was that I could do something like this:

```ts
export const { getClient } = registerApolloClient((previewToken: PreviewToken | null) => {
  return new ApolloClient({
    link: authLink.concat(
      createHttpLink({
        uri: `my endpoint based on values from ${previewToken}`,
      })
    ),
    cache: new InMemoryCache(),
  });
});
```

But I'm getting

```
Argument of type '(previewToken: PreviewToken | null) => ApolloClient<NormalizedCacheObject>' is not assignable to parameter of type '() => ApolloClient<any>'.
  Target signature provides too few arguments. Expected 1 or more, but got 0.
```

Plus I don't know whether the underlying implementation would correctly keep the different clients separate.

tremby:
This allows the user to have separate clients based on arguments passed, in order to achieve things such as content previews.

Closes #67

### DIFF
--- a/package/README.md
+++ b/package/README.md
@@ -71,6 +71,17 @@ You can then use that `getClient` function in your server components:
 const { data } = await getClient().query({ query: userQuery });
 ```
 
+#### Multiple clients
+
+If you need to manage multiple clients, for example if you have different endpoints for content previews,
+the function you pass to `registerApolloClient` can accept arbitrary arguments.
+If you're using Typescript, you can hint their types as a tuple:
+
+```ts
+export const { getClient } = registerApolloClient<[string | null]>((previewToken: string | null) => {
+  ...
+});
+
 ### In SSR
 
 If you use the `app` directory, each Client Component _will_ be SSR-rendered for the initial request. So you will need to use this package.
--- a/package/README.md
+++ b/package/README.md
@@ -79,7 +79,7 @@ If you're using Typescript, you can hint their types as a tuple:
 
 ```ts
 export const { getClient } = registerApolloClient<[string | null]>((previewToken: string | null) => {
-  ...
+  // ...
 });
 
 ### In SSR
--- a/package/src/rsc/registerApolloClient.tsx
+++ b/package/src/rsc/registerApolloClient.tsx
@@ -11,7 +11,7 @@ function assertRSC(
   }
 }
 
-export function registerApolloClient(makeClient: () => ApolloClient<any>) {
+export function registerApolloClient<Args extends any[] = any[]>(makeClient: (...args: Args) => ApolloClient<any>) {
   assertRSC(React);
   const getClient = React.cache(makeClient);
   return {


package/README.md
```diff
@@ -71,6 +71,17 @@ You can then use that `getClient` function in your server components:
...
+
+If you need to manage multiple clients, for example if you have different endpoints for content previews,
+the function you pass to `registerApolloClient` can accept arbitrary arguments.
+If you're using Typescript, you can hint their types as a tuple:
+
+```ts
+export const { getClient } = registerApolloClient<[string | null]>((previewToken: string | null) => {
+  ...
+});
+
```
jerelmiller, line 84:
````suggestion
```ts
export const { getClient } = registerApolloClient<[string | null]>((previewToken: string | null) => {
  ...
});
```
````

It looks like this code block is missing the closing ticks, so the formatting is a bit weird right now 😆. Here is the [rendered README](https://github.com/tremby/apollo-client-nextjs/tree/register-client-allow-args) for reference.

File: package/README.md [hunk 2]
```diff
@@ -71,6 +71,17 @@ You can then use that `getClient` function in your server components:
...
 
+#### Multiple clients
+
+If you need to manage multiple clients, for example if you have different endpoints for content previews,
+the function you pass to `registerApolloClient` can accept arbitrary arguments.
+If you're using Typescript, you can hint their types as a tuple:
+
+```ts
+export const { getClient } = registerApolloClient<[string | null]>((previewToken: string | null) => {
+  ...
```
jerelmiller, line 82:
```suggestion
  // ...
```

I always prefer to use a comment like this to denote a code placeholder so that it maintains proper syntax and code highlighting. Would you mind switching over to something like this?

package/README.md
```diff
@@ -71,6 +71,17 @@ You can then use that `getClient` function in your server components:
...
 ```
 
+#### Multiple clients
+
+If you need to manage multiple clients, for example if you have different endpoints for content previews,
+the function you pass to `registerApolloClient` can accept arbitrary arguments.
+If you're using Typescript, you can hint their types as a tuple:
+
+```ts
+export const { getClient } = registerApolloClient<[string | null]>((previewToken: string | null) => {
```
jerelmiller, line 81:
```suggestion
export const { getClient } = registerApolloClient((token: string | null) => {
```

Couple things here:

1. Let's rely on type inference where possible since its easier to understand than passing the tuple as a generic. We find a lot of people will copy/paste code, so we tend to prefer showing off the "optimal" code when possible. I've pulled this down locally and this change works as expected.
2. The name `previewToken` feels a bit specific for the README. Could we make this a tad more generic to something like `token` or `authToken` which are more wide-spread?

package/README.md
```diff
@@ -71,6 +71,17 @@ You can then use that `getClient` function in your server components:
 const { data } = await getClient().query({ query: userQuery });
 ```
 
+#### Multiple clients
+
+If you need to manage multiple clients, for example if you have different endpoints for content previews,
```
jerelmiller, line 76: To be honest, the "for example" feels a little too specific. I'm not sure most apps would have a "content preview", nor do I think many people might understand what this is (to be honest I'm one of those 😅).

Instead, could we use either something like an auth token, or a versioned endpoint or something as an example? This might resonate with readers a bit more. I'm also for removing this part of the sentence altogether if we can't come up with a great example.

package/README.md
```diff
@@ -71,6 +71,17 @@ You can then use that `getClient` function in your server components:
...
+the function you pass to `registerApolloClient` can accept arbitrary arguments.
+If you're using Typescript, you can hint their types as a tuple:
```
jerelmiller, line 78:
```suggestion
you can pass a function to `registerApolloClient` that accepts arguments.
```

I'd suggest removing that last sentence. I think the code sample does a good job showing off how to use TypeScript with this function, so this point is inferred. I also had a suggestion for a slight tweak in that other sentence that gives it more active voice.

package/README.md
jerelmiller: As part of this code block, I'd also love to show off _how_ you'd use the argument to ensure people understand that `getClient` is the function affected by the argument. Stealing from the example in the section before this, I'd suggest adding this line:

```ts
const { data } = await getClient(token).query({ query: userQuery });
```

jerelmiller (commented):
The change itself looks great to me! Most of my suggestions were related to the documentation of it. I'd love to have @phryneas take a look as well and give the stamp of approval before we get this merged.

phryneas:
I'm sorry, but to be honest, I'm not really enthusiastic about this idea - I see this opening up situations with a lot of potential for user error.

The point of `getClient()` right now is that it returns the same client to you every time you call it within a single RSC render pass.

If you wanted two fundamentally different clients, you could already do

```js
const { getClient: getFirstClient } = registerApolloClient(makeClient1)
const { getClient: getSecondClient } = registerApolloClient(makeClient2)
```

and generally, you probably already want to capture something like a `token` in scope at that point already - otherwise you would have to pass that `token` into *all* of your components, while avoiding to use any kind of global variable, since those would be shared between multiple requests rendering at the same time. (Big footgun here!)

If you can't do this on initialization, you can still pass the `token` into the link chain using `context` when calling `query`:
```js
getFirstClient().query({ query, variables, context: { token }})
```

tremby:
I've responded over in #67. Thanks for taking a look.

PR #69, opened 2023-07-21
README: Fix js to jsx for SSR

jeffng50:
### Why is the change needed
Changing the README to specify that we should use the filename `app/ApolloWrapper.jsx` instead of `app/ApolloWrapper.js` in the _**In SSR**_ section, because there is a React component in there, and we will get `Parsing error: '>' expected eslint` error when it's in `.js` (EDIT: correction it'll only error in Typescript files)

apollowrapper
I am using Typescript in this file

### What I have changed
- `.js` to `.jsx` in the _**In SSR**_ section

### DIFF
--- a/package/README.md
+++ b/package/README.md
@@ -75,7 +75,7 @@ const { data } = await getClient().query({ query: userQuery });
 
 If you use the `app` directory, each Client Component _will_ be SSR-rendered for the initial request. So you will need to use this package.
 
-First, create a new file `app/ApolloWrapper.js`:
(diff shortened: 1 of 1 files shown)


jeffng50:
Correction: Just figured out the error is happening only on Typescript and not Javascript, so I was wrong. However it would be beneficial to state? I'll let y'all decide

phryneas:
Yeah, it only makes a difference for TypeScript files - but you're right, it definitely cannot hurt :)

Thanks for the contribution!

PR #70, opened 2023-07-21
fix: tag snapshot npm releases with snapshot tag instead of latest

alessbell:
Just noticed a snapshot release was tagged with `latest` when I just `npm i`'d the library. We don't need every snapshot release to have its own tag (this list would grow infinitely on our npm page while providing basically no utility), so I went with the generic `snapshot` tag that will be shared across all snapshot releases.


jerelmiller (approved):

phryneas:
Thanks, I didn't know that npm tagged `latest` by default!

Issue #72, opened 2023-08-01
Support Next.js cache options on `fetch`


moshest:
I follow the tutorial of [using Apollo with Next.js App directory](https://www.apollographql.com/blog/announcement/frontend/using-apollo-client-with-next-js-13-releasing-an-official-library-to-support-the-app-router/).

I notice the tutorial ask to add the `force-dynamic` parameter to the page to control if Next.js should cache the response. Although this solution works I find it very hard to maintain, especially if I have child components that require dynamic rendering.

Next.js solve that by providing `cache: 'no-store'` on the `fetch` function ([docs](https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating)) so the engine will know which request shouldn't be cached.

I currently solve it for my use case by setting it to `no-store` for all GraphQL requests:

```ts
export const link = new HttpLink({
  uri: `https://example.com/graphql`,
  fetchOptions: {
    cache: 'no-store',
  },
});
```

Is there an option to control the `fetch` option using the `fetchPolicy` on `useSuspenseQuery`?

I expect `useSuspenseQuery` to send `no-store` for every request UNLESS I choose to cache it on build time. Something like the following will be ideal:

```ts
// default behavior
const { data } = useSuspenseQuery(MY_QUERY, {
  // fetchCache: 'no-store',
});

// override default
const { data } = useSuspenseQuery(MY_QUERY, {
  fetchCache: 'force-cache',
});
```

phryneas:
That said, that's pretty much baked in already with what you have.

```js
export const link = new HttpLink({
  uri: `https://example.com/graphql`,
  fetchOptions: {
    cache: 'no-store',
  },
});
```

sets that default, and you can overwrite it with
```js
// override default
const { data } = useSuspenseQuery(MY_QUERY, {
  context: { fetchOptions: { fetchCache: "force-cache" }},
});
```

moshest:
You are right, I manage to config it as following:

```ts
// set default cache options for all requests
const link = new HttpLink({
  uri: `https://example.com/graphql`,
  fetchOptions: {
    cache: 'no-store',
  },
});

// override default
const { data } = useSuspenseQuery(MY_QUERY, {
  context: { fetchOptions: { cache: "force-cache" }},
});
```

[...]

PR #73, opened 2023-08-02
Update README.md to show how to override default fetchOptions

phryneas:
See https://github.com/apollographql/apollo-client-nextjs/issues/72#issuecomment-1661653844

### DIFF
--- a/package/README.md
+++ b/package/README.md
@@ -97,6 +97,8 @@ function makeClient() {
     // you can disable result caching here if you want to
     // (this does not work if you are rendering your page with `export const dynamic = "force-static"`)
     fetchOptions: { cache: "no-store" },
+    // you can override the default set here for individual queries in your components by calling them like
+    // const { data } = useSuspenseQuery(MY_QUERY, { context: { fetchOptions: { cache: "force-cache" }}});
   });
 
   return new NextSSRApolloClient({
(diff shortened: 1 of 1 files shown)


moshest (approved):

package/README.md
```diff
@@ -97,6 +97,8 @@ function makeClient() {
     // you can disable result caching here if you want to
     // (this does not work if you are rendering your page with `export const dynamic = "force-static"`)
     fetchOptions: { cache: "no-store" },
+    // you can override the default set here for individual queries in your components by calling them like
+    // const { data } = useSuspenseQuery(MY_QUERY, { context: { fetchOptions: { cache: "force-cache" }}});
```
alessbell, line 101: Some minor nits: calling out this goes in `context` so no one misses that, and should is `to an Apollo Client data fetching hook` or `to an Apollo Client suspense hook` more correct here? I went with the former because I thought this might still be relevant to `useQuery` calls.

```suggestion
    // you can override the default `fetchOptions` on a per query basis
    // via the `context` property on the options passed as a second argument
    // to an Apollo Client data fetching hook, e.g.:
    // const { data } = useSuspenseQuery(MY_QUERY, { context: { fetchOptions: { cache: "force-cache" }}});
```

alessbell (approved):
Minor wordsmithing question/comment but LGTM 🙌

PR #75, opened 2023-08-03
chore: remove experimental flag as it's stable now

ScottAgirs:
I wasn't able to properly test, but on my own separate Next.js instance with the same npm versions works flawflessly without the `appDir: true` flag.

As seen enclosed in the "Good to know" callout:
https://nextjs.org/docs/app/api-reference/next-config-js/appDir


phryneas:
Good call, thank you!

PR #90, opened 2023-08-30
Use server-only package instead of custom environment detection

leonchabbey:
I saw this issue when trying to fetch data from the `sitemap.ts` file using the `getClient` / `registerApolloClient` (like it's done for server only code).
Apparently the "client version" of React is somehow available when the sitemap is getting generated which was triggering the custom detection currently implemented.

I tried using `import 'server-only` instead as it is the recommended way to make sure something isn't imported from a client component and all cases seem to work:
- From CC -> crash (✅ )
- From RSC -> ✅ 
- From sitemap.ts -> ✅

### DIFF
--- a/package/package.json
+++ b/package/package.json
@@ -64,6 +64,7 @@
     "react": "^18"
   },
   "dependencies": {
+    "server-only": "^0.0.1",
     "superjson": "^1.12.2",
     "ts-invariant": "^0.10.3"
   }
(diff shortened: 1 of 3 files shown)


phryneas:
> Apparently the "client version" of React is somehow available when the sitemap is getting generated which was triggering the custom detection currently implemented.

That would be very worrying as that client version has no working version of `React.cache` 🤔 

Apart form that: this is a great change, and will also play nicely with https://github.com/apollographql/apollo-client/pull/11175 - thank you!

I just wonder why CI didn't kick off?

phryneas:
And released in https://github.com/apollographql/apollo-client-nextjs/releases/tag/v.0.4.2 :)

leonchabbey:
Great! Thanks a lot for being so quick getting this merged 👏

PR #117, opened 2023-10-28
Update `next` peerDependency to include 14.0.0

none23:
Since `next@14.0.0` is out, and seems to be working fine with this package, I've updated the package's `peerDependencies` to reflect that.

 I've also updated the examples to use the newer version


phryneas:
I'm so sorry I actually missed this!

By now, I've done the same, and it has been released in the `0.5.x` version range.

Thank you again for the PR, I would have loved to merge this in instead, but I really didn't see it before.

PR #144, opened 2023-11-30
prepare for change in `QueryManager.inFlightLinkObservable` (Apollo Client 3.9 beta)

phryneas:
See changes in https://github.com/apollographql/apollo-client/pull/11345

Unfortunately, we already bumped the dependencies for the alphas/betas/rcs, so we can't just force an update that way.


phryneas:
See https://github.com/apollographql/apollo-client-nextjs/actions/runs/7047409099?pr=144 
This is the error we are looking for, that's triggered on the Apollo Client side:
image


And now let's make this work :)

phryneas:
Working - requesting a review :)

jerelmiller (approved):

Issue #126, opened 2023-11-09
Large bundle size compared to the standard @apollo/client


tombryden:
Hi,

I have noticed when running `npm run build` the first load JS is incredibly large when using any of the hooks from this library when compared with the standard `@apollo/client`

### Import size using 'Import Cost' extension of both the libraries:
![image](https://github.com/apollographql/apollo-client-nextjs/assets/31113935/2a246e19-c5a5-4024-a205-32be904f0d2f)
vs
![image](https://github.com/apollographql/apollo-client-nextjs/assets/31113935/1acda34e-da92-4743-991e-1ab48a18da73)

### npm run build results
![image](https://github.com/apollographql/apollo-client-nextjs/assets/31113935/73e67f7c-97ee-4170-a64c-b33ef53509f7)
vs
![image](https://github.com/apollographql/apollo-client-nextjs/assets/31113935/86cf03eb-f894-45d5-9180-21c6f84062e3)

Thanks

phryneas:
I've seen something similar in #95 where using `NextSSRApolloClient` instead of `ApolloClient` in `registerApolloClient` (so for RSC, not SSR) drove bundle size a lot - is that maybe happening here?

tombryden:
I am using the NextSSRApolloClient, however I just tested with a fresh project and even without any wrapper the package size is 75+kb for any import from `@apollo/experimental-nextjs-app-support/ssr`.

Fresh project, only containing `page.tsx` and `layout.tsx` causes the same issue:

```
"use client";

import { useQuery } from "@apollo/experimental-nextjs-app-support/ssr";

console.log(useQuery);

export default function Home() {
  return (
    <main>
    </main>
  );
}
```

tombryden:
https://github.com/apollographql/apollo-client-nextjs/issues/95#issuecomment-1721965647 - I have also noticed strange behaviour in other instances with bundle sizes.

Just raised a bug with MUI earlier today showing similar behaviour - https://github.com/mui/material-ui/issues/39814

tombryden:
@phryneas I have just re-tested this on 14.0.4 but still the same issue. The MUI bug I mentioned above appears to be fixed on 14.0.4 so I had some hope

phryneas:
Looking into this, this looks like some bug in the next.js build report.

If I build the pages (working with our integration test here), I get these entries in my manifest (`.next/app-build-manifest.json`):

```json
{
  "/cc/dynamic/useSuspenseQuery/page": [
[...]

PR #161, opened 2024-01-03
reduce bundle size by adjusting imports

phryneas:
This should especially tree-shake a large part of the `graphql` package that showed up in https://github.com/apollographql/apollo-client-nextjs/issues/126#issuecomment-1875590159

### DIFF
--- a/package/src/ssr/NextSSRApolloClient.tsx
+++ b/package/src/ssr/NextSSRApolloClient.tsx
@@ -9,7 +9,7 @@ import {
   Observable
 } from "@apollo/client";
 import type { QueryManager } from "@apollo/client/core/QueryManager";
-import { print } from "graphql";
+import { print } from "@apollo/client/utilities";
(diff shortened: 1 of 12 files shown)


package/src/ssr/NextSSRApolloClient.tsx
```diff
@@ -1,16 +1,17 @@
...
-  Observable,
   FetchResult,
-  DocumentNode,
+  DocumentNode} from "@apollo/client";
+import {
+  ApolloClient,
+  Observable
 } from "@apollo/client";
 import type { QueryManager } from "@apollo/client/core/QueryManager";
-import { print } from "graphql";
```
phryneas, line 11: This is what apparently has the big bundle size impact, as for some reason that's not tree-shaken by Next.js otherwise.

phryneas:
Released in https://github.com/apollographql/apollo-client-nextjs/releases/tag/v.0.6.0

PR #180, opened 2024-01-24
feat: adds user survey workflow and link in README

alessbell:


travigd:
FYI @alessbell looks like the survey isn't actually publicly available:

> You need permission
> This form can only be viewed by users in the owner's organization.

alessbell:
@travigd oops! Thanks for the heads up! I've just pinged my colleague - this should be fixed today, at which point we'd still very much appreciate your feedback :)

calvincestari:
Hi @travigd the link is fixed. Thank you very much for letting us know.

PR #187, opened 2024-02-02
Removing confusing installation language in README.md

spenweb:
Removed unnecessary mention of the rc version of the @apollo/client package. This should remove confusion around which version should be installed. See https://github.com/apollographql/apollo-client-nextjs/releases/tag/v.0.8.0


phryneas:
Good catch, thank you! :)

PR #193, opened 2024-02-12
split up into two packages

phryneas:


.github/workflows/publish.yml
```diff
@@ -1,8 +1,11 @@
 name: Publish Package to npmjs
 on:
-  release:
-    types: [published]
   workflow_dispatch:
+    inputs:
+      version:
+        description: "The version to publish"
+        required: true
```
phryneas, line 7: This workflow now needs to be manually triggered from the Github Actions UI, as we need to be very specific about the version that will be applied to all packages (and parsing some release title for a version number might turn out flimsy).

This also allows us to publish under a certain tag.

.github/workflows/snapshot-release.yml
```diff
@@ -1,5 +1,6 @@
 name: Snapshot Release
 on:
+  workflow_dispatch:
```
phryneas, line 3: This workflow can now also be triggered from the GitHub UI, which has the benefit that it will actually run the workflow version of that branch and not the workflow from `main` - meaning, we can not test it before merging into `main`.

packages/client-react-streaming-support/src/DataTransportAbstraction/useTransportValue.tsx
```diff
@@ -1,5 +1,5 @@
 "use client";
-import { useContext, useEffect, useState } from "rehackt";
+import { useContext, useEffect, useState } from "react";
```
phryneas, line 2: With all the exports conditions we have in play for these packages now, we don't need `rehackt` anymore.

.github/workflows/publish.yml
jerelmiller: Would it be worth spending some effort getting changesets integrated? It handles workspaces with dependent packages pretty well from what I've seen (check our brand repo for an example, specifically the tailwind package that depends on the base `brand` package)

packages/client-react-streaming-support/package.json
jerelmiller, line 1: [nit] Should the workspace name match the package name? In other words, would it be alright to drop the `-support` suffix from this folder name?

jerelmiller (approved):
💯 🔥 💯 Nice work!

PR #196, opened 2024-02-14
add parameter type support to getClient

rezof:
I kept getting `getClient Expected 0 arguments, but got ...` because makeClient has no parameters

### DIFF
--- a/package/src/rsc/registerApolloClient.tsx
+++ b/package/src/rsc/registerApolloClient.tsx
@@ -3,7 +3,7 @@ import "server-only";
 import type { ApolloClient } from "@apollo/client";
 import * as React from "react";
 
-export function registerApolloClient(makeClient: () => ApolloClient<any>) {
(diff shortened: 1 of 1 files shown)


phryneas:
That is kinda intentional - if I may ask, what are you passing in there?

rezof:
@phryneas i pass `fetchOptions:{ cache: ...}` for next as well as fetch headers. 

I just remembered that i can use context for setting such things. the only difference is they have to be provided to every query/mutation on page while `getClient`  passes them directly to the `HttpLink`

If you want to keep it without parameters you can close the PR.

Thanks 🙌

phryneas:
Yeah, you definitely should avoid doing it this way.

If you pass different arguments into `getClient` for two calls, you will get two different client instances. 

The purpose of `getClient` is to create a new Apollo Client instance for every new request, but to always return the same instance within that request.

As soon as you call it with potentially different arguments, that functionality stops working.

=> Yeah, please use the per-request context for this :)

I'm going to close this issue. Have a great day!

PR #203, opened 2024-02-19
0.10.0 (1) - Throw error when passing arguments into `getClient`.

phryneas:
It has come again and again that people are trying to do this, so we should be more strict about the fact that this is not supported.

Happy for any wording suggestions.


packages/client-react-streaming/src/registerApolloClient.tsx
```diff
@@ -2,7 +2,21 @@ import type { ApolloClient } from "@apollo/client/index.js";
...
-  const getClient = cache(makeClient);
+  function wrappedMakeClient() {
+    if (arguments.length) {
+      throw new Error(
+        `
+You cannot pass arguments into \`getClient\`!
+If you were to pass arguments into \`getClient\`, you would get a different copy
+of Apollo Client for every time you called it with different arguments, resulting
+in duplicate requests and a non-functional cache, essentially defeating the purpose
+of \`registerApolloClient\` and \`getClient\`.
```
jerelmiller, line 13:
```suggestion
You cannot pass arguments into \`getClient\`.
Passing arguments to \`getClient\` returns a different copy
of Apollo Client each time its called with different arguments, potentially 
resulting in duplicate requests and a non-functional cache. This defeats 
the purpose of \`registerApolloClient\` and \`getClient\`.
```

I really like this change! I'd word this a little more using the present tense rather than future tense.

I'd also consider dropping the last part altogether since risks stated before it provide enough information on why this isn't allowed.

```suggestion
You cannot pass arguments into \`getClient\`!
If you were to pass arguments into \`getClient\`, you would get a different copy
of Apollo Client for every time you called it with different arguments, resulting
in duplicate requests and a non-functional cache.
```

jerelmiller (commented):
Another solid, quality of life improvement!

packages/client-react-streaming/src/registerApolloClient.tsx
phryneas:
```suggestion
You cannot pass arguments into \`getClient\`.
Passing arguments to \`getClient\` returns a different copy
of Apollo Client each time its called with different arguments, potentially 
resulting in duplicate requests and a non-functional cache. 
```

packages/client-react-streaming/src/registerApolloClient.tsx
```diff
@@ -2,7 +2,20 @@ import type { ApolloClient } from "@apollo/client/index.js";
...
 export function registerApolloClient(makeClient: () => ApolloClient<any>) {
-  const getClient = cache(makeClient);
+  function wrappedMakeClient() {
+    if (arguments.length) {
+      throw new Error(
+        `
+You cannot pass arguments into \`getClient\`!
+If you were to pass arguments into \`getClient\`, you would get a different copy
+of Apollo Client for every time you called it with different arguments, resulting
+in duplicate requests and a non-functional cache.
```
phryneas, line 12:
```suggestion
You cannot pass arguments into \`getClient\`.
Passing arguments to \`getClient\` returns a different instance
of Apollo Client each time it is called with different arguments, potentially 
resulting in duplicate requests and a non-functional cache. 
```

jerelmiller (approved):
🎉

jerelmiller (approved):
💯

phryneas:
Uhh, wtf. deleted the parent branch and now this is closed 🤦

Issue #195, opened 2024-02-14
async registerApolloClient ?


Dakuan:
Be really handy if `registerApolloClient` was async...here's what im trying to do

```
import { HttpLink } from "@apollo/client";
import {
  NextSSRInMemoryCache,
  NextSSRApolloClient,
} from "@apollo/experimental-nextjs-app-support/ssr";
import { registerApolloClient } from "@apollo/experimental-nextjs-app-support/rsc";
import { runWithAmplifyServerContext } from "../amplify/amplifyServerUtils";
import { fetchAuthSession } from "aws-amplify/auth/server";
import { cookies } from "next/headers";

export const { getClient } = registerApolloClient(async () => {
  const token = await runWithAmplifyServerContext({
    nextServerContext: { cookies },
    operation: async (contextSpec) => {
      const session = await fetchAuthSession(contextSpec);
      return session.tokens?.accessToken.toString();
    },
  });
  const httpLink = new HttpLink({
    uri: "http://localhost:4000/graphql",
    headers: {
      Authorization: `Bearer ${token}`,
    },
  });

  return new NextSSRApolloClient({
    cache: new NextSSRInMemoryCache(),
    link: httpLink,
  });
});
```

all my auth its handled by aws congnito / amplify. I can get my jwts using the amplify utils, and them pass them to the API. But, its an async process, so needs to live in an async fn...

Dakuan:
this is my workaround, ugly but works:

```typescript
import { HttpLink } from "@apollo/client";
import {
  NextSSRInMemoryCache,
  NextSSRApolloClient,
} from "@apollo/experimental-nextjs-app-support/ssr";
import { registerApolloClient } from "@apollo/experimental-nextjs-app-support/rsc";
import { runWithAmplifyServerContext } from "../amplify/amplifyServerUtils";
import { fetchAuthSession } from "aws-amplify/auth/server";
import { cookies } from "next/headers";

export const getClient = async () => {
  const token = await runWithAmplifyServerContext({
    nextServerContext: { cookies },
    operation: async (contextSpec) => {
      const session = await fetchAuthSession(contextSpec);
      return session.tokens?.accessToken.toString();
    },
  });
  const { getClient } = registerApolloClient(() => {
    const httpLink = new HttpLink({
      uri: "http://localhost:4000/graphql",
      headers: {
        Authorization: `Bearer ${token}`,
      },
    });
    return new NextSSRApolloClient({
[...]

PR #204, opened 2024-02-19
`registerApolloClient`: allow for async `makeClient` function

phryneas:
This was requested over in #195, so we could consider doing this.


jerelmiller (approved):
I see no reason why we shouldn't allow this. Nice! 🎉

PR #205, opened 2024-02-19
0.10.0 (2) - Warn if makeClient returns the same client for multiple requests.

phryneas:


packages/client-react-streaming/src/registerApolloClient.tsx
```diff
@@ -1,8 +1,24 @@
...
 export function registerApolloClient(makeClient: () => ApolloClient<any>) {
-  function wrappedMakeClient() {
+  // this function will always create a new object
+  // so we can warn if `cachedMakeWrappedClient` creates a new "wrapper",
+  // but with a `client` property that we have already seen before.
+  // In that case, not every call to `makeClient` would create a new
+  // `ApolloClient` instance.
+  function makeWrappedClient() {
+    return { client: makeClient() };
+  }
```
jerelmiller, line 17: I think I got a little confused at first on why you'd need the wrapper object here instead of just returning `makeClient` directly. Looking at the [React documentation](https://react.dev/reference/react/cache#noun-labs-1201738-(2)), I see this point:

> React will invalidate the cache for all memoized functions for each server request.

I'm assuming this is the reason you're returning a wrapping object rather than the client directly? Is it just harder to detect unique references to `client` otherwise?

If so, perhaps you can update the comment to provide a little more clarity. Something like:

```suggestion
  // React invalidates the cache on each server request, so the wrapping
  // object is needed to properly detect whether the client is a unique 
  // reference or not. We can warn if `cachedMakeWrappedClient` creates a new "wrapper",
  // but with a `client` property that we have already seen before.
  // In that case, not every call to `makeClient` would create a new
  // `ApolloClient` instance.
  function makeWrappedClient() {
    return { client: makeClient() };
  }
```

Do let me know if I'm thinking about this in the wrong way.

jerelmiller (commented):
I assume that you haven't written tests because this is an [idea] right? 

In general I like this idea a lot since it at least gives some indication of a potential footgun for those that are possibly unaware of the impact of returning the same client instance. Nice work!

phryneas:
Yup, I wanted to make sure we actually want to do this before I get into the nitty-gritty of figuring out how to write a test against a RSC environment :)

packages/client-react-streaming/src/registerApolloClient.tsx
phryneas: The user could (and should!) call the memoized `getClient` function a dozen times while working on a request.

So it is normal that this memoized function runs multiple times and returns a `client` instance we have seen before. We need one piece of additional information: "Was this the first time for this request we called `getClient`?" Because only in this case, returning a "known" client instance would be problematic.

So we do something where we **know** that it will be created once per request uniquely - we memoize creating an object (the wrapper).

If we see a new object and a new client - all good.
If we see a known object and a known client - all good.
If we see a new object, but an old client - this is bad, a user is not creating a new client in their `makeClient` function.

phryneas: => Yeah, I think your comment makes a ton of sense - taking it!

packages/client-react-streaming/src/registerApolloClient.tsx
```diff
@@ -10,20 +17,50 @@ export function registerApolloClient(makeClient: () => ApolloClient<any>): {
...
-    return makeClient();
+    const wrapper = cachedMakeWrappedClient();
+    if (seenWrappers && seenClients) {
+      if (!seenWrappers.has(wrapper)) {
+        if (seenClients.has(wrapper.client)) {
+          console.warn(
+            `
+Multiple calls to \`getClient\` for different requests returned the same client instance.
+This means that private user data could accidentally be shared between requests.
+This for example happens if you create a global \`ApolloClient\` instance and your \`makeClient\`
```
jerelmiller, line 51:
```suggestion
This happens, for example, if you create a global \`ApolloClient\` instance and your \`makeClient\`
```

packages/client-react-streaming/src/registerApolloClient.tsx
```diff
@@ -10,20 +17,50 @@ export function registerApolloClient(makeClient: () => ApolloClient<any>): {
...
+implementation just looks like \`() => client\`.
+Please insure to always call \`new ApolloClient\` **inside** your \`makeClient\` function and to
```
jerelmiller, line 53:
```suggestion
Always call \`new ApolloClient\` **inside** your \`makeClient\` function and
```

jerelmiller (approved):
Had some small suggestions on wording, but this is great!

PR #206, opened 2024-02-19
[idea] ensure that wrapped hooks are called instead of their upstream counterparts

phryneas:
This would work with `useFragment`, `useQuery`, `useSuspenseQuery` and `useBackgroundQuery` since all of those call either `watchQuery` or `watch`, and we can overwrite these.

It does not work with `useReadQuery` - @jerelmiller do you have an idea how we could make it work there?


phryneas:
Closing in favor of #216

PR #215, opened 2024-02-22
[idea] wrap upstream package hooks with streaming functionality

phryneas:
This, in combination with https://github.com/apollographql/apollo-client/pull/11615 would make it possible to inject functionality into the `@apollo/client` hooks, so it wouldn't matter if users used the `@apollo/client` hooks, or the hooks from this package.

That has one clear downside, though: 
This is a side effect that pulls all the hooks into the bundle - tree shaking of the hooks would become impossible. 

It might very well be worth it, though 🤔 

As an alternative, we could export `set up` functions for all the hooks and introduce people to execute those set up functions, which would circumvent that, but creates more friction.

### DIFF
--- a/packages/client-react-streaming/src/DataTransportAbstraction/hooks.ts
+++ b/packages/client-react-streaming/src/DataTransportAbstraction/hooks.ts
@@ -1,43 +1,68 @@
 "use client";
+import type { ApolloClient } from "@apollo/client/index.js";
 import {
-  useFragment as orig_useFragment,
-  useSuspenseQuery as orig_useSuspenseQuery,
-  useReadQuery as orig_useReadQuery,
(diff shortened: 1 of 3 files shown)


packages/client-react-streaming/src/DataTransportAbstraction/hooks.ts
```diff
@@ -1,43 +1,68 @@
...
+};
+
+wrap(
+  useFragment,
+  ["data", "complete", "missing"],
+  (options) => options?.client
+);
+wrap<typeof useQuery>(
   process.env.REACT_ENV === "ssr"
     ? (query, options) =>
```
phryneas, line 30: This doesn't work like this yet, I'll have to take another look.

PR #216, opened 2024-02-23
0.9.0 (2) - wrap hook functionality from ApolloClient instance

phryneas:
This PR builds on https://github.com/apollographql/apollo-client/pull/11617 and does the same as #215 or #206, but should tree-shake better than #215 and has better UX than #206


packages/client-react-streaming/src/DataTransportAbstraction/index.ts
```diff
@@ -7,7 +7,7 @@ export {
   useSuspenseQuery,
   useReadQuery,
   useBackgroundQuery,
-} from "./hooks.js";
+} from "@apollo/client/index.js";
```
phryneas, line 10: No more custom implementations & it doesn't matter where it's imported from 🎉

jerelmiller: I'm curious, is there any reason not to just deprecate these exports and have users import directly from `@apollo/client`? That could simplify things even further 🙂

packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx
```diff
@@ -26,13 +27,24 @@ type SimulatedQueryInfo = {
...
     super(options);
+
+    if (!(this.cache instanceof WrappedInMemoryCache)) {
+      throw new Error(
+        "When using Apollo Client streaming SSR, you must use the `InMemoryCache` variant provided by the streaming package."
+      );
+    }
   }
 
+  [wrappers] = hookWrappers;
```
jerelmiller, line 42: See https://github.com/apollographql/apollo-client/pull/11617/files#r1500875764

I'd be curious if we could store this on `QueryManager` which would make a path for `useReadQuery` to work in all scenarios.

jerelmiller (approved):
I like this a lot! Really creative solution to allow for more simplification.

packages/client-react-streaming/src/DataTransportAbstraction/index.ts
phryneas: Removing it from this package, but keeping it in the "outer" package for now. (Will drop it in a future version.)

jerelmiller (approved):

jerelmiller (approved):

PR #219, opened 2024-02-28
0.9.0 (3) - add integration test with experimentally patched React version

phryneas:
This adds an integration test for the "experimental React hooks" solution and moves the `@apollo/client-react-streaming/experimental-react-transport` out of the package, only into the test.

It's not overly complicated, and we probably don't *need* to really ship it as part of a package.


integration-test/experimental-react/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@integration-test/experimental-react",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "prepare": "rm -rf node_modules/@apollo/client-react-streaming; mkdir node_modules/@apollo/client-react-streaming && cp -r ../../packages/client-react-streaming/{package.json,dist} node_modules/@apollo/client-react-streaming",
```
phryneas, line 7: This is necessary so it doesn't pick up the `react` folder from `packages/client-react-streaming/node_modules` and mixes two react versions.

integration-test/experimental-react/src/WrappedApolloProvider.tsx
```diff
@@ -1,16 +1,17 @@
...
-  useRef,
-} from "react";
+import {
+  WrapApolloProvider,
+  DataTransportContext,
+} from "@apollo/client-react-streaming";
 import type { DataTransportProviderImplementation } from "@apollo/client-react-streaming";
-import { DataTransportContext } from "@apollo/client-react-streaming";
+import { useMemo, useActionChannel, useStaticValue, useRef } from "react";
+
```
plausible-phry, line 14: Note to self: double check the approach of queueing a single promise for request & response

phryneas: https://github.com/apollographql/apollo-client-nextjs/pull/223

integration-test/experimental-react/server.js
```diff
@@ -0,0 +1,101 @@
...
+const ABORT_DELAY = 10000;
+
+// Create http server
+const app = express();
+
+// Add Vite or respective production middlewares
+let vite;
+let bootstrapModules = [];
+let assets = [];
+console.log({ isProduction });
```
jerelmiller, line 18:
```suggestion
```

Not sure if this was meant to stay or not 🙂

integration-test/experimental-react/server.js
```diff
@@ -0,0 +1,101 @@
...
+    /<script type="module" \w+ src="(.*)">/g
+  )) {
+    bootstrapModules.push(script[1]);
+  }
+  for (const link of index.matchAll(
+    /<link rel="stylesheet" \w+ href="(.*)">/g
+  )) {
+    assets.push(link[1]);
+  }
+}
```
jerelmiller, line 43: I know this is meant to be more of a "demo" (ish), but any chance we could flip the conditional here? I always find it easier with `if/else` to use the positive connotation first.

Rather than:
```ts
if (!isProduction) {
  // other envs
} else {
  // production builds
}
```

Use this:
```ts
if (isProduction) {
  // production stuff
} else {
  // other environments
}
```

This keeps the "production" configuration closer to the conditional test for it, rather than breaking it apart.

integration-test/experimental-react/server.js
```diff
@@ -0,0 +1,101 @@
...
+      setImmediate(() => pipe(res));
+    },
+    onShellError(x) {
+      console.log("Shell error", x);
+      // Something errored before we could complete the shell so we emit an alternative shell.
+      res.statusCode = 500;
+      res.send("<!doctype><p>Error</p>");
+    },
+    onError(x) {
+      console.log("Error", x);
```
jerelmiller, line 84:
```suggestion
```

Since you already have a `console.error` below, could we just use that?

jerelmiller (approved):
My comments were petty 😆 , otherwise looks great! 🎉

integration-test/experimental-react/server.js
phryneas: This is based on the [Vite React SSR template](https://github.com/bluwy/create-vite-extra/blob/master/template-ssr-react/server.js#L22) and I'm trying to keep the diff as small as possible, so while I agree and would really like to swap it, it's probably better we don't :)

integration-test/experimental-react/server.js
phryneas: Uh yeah, I'm generally gonna remove a bunch of these logs :)

integration-test/experimental-react/server.js
jerelmiller: I didn't realize these were copy-pasted so makes sense! Thanks for pointing me to that.

PR #221, opened 2024-03-01
0.9.0 (6) - Add integration test for vitest

phryneas:
Vitest is a bit tricky, both because we have to manually set `--conditions browser` and somehow get around the dual package hazard. This seems to work as a minimal setup.


integration-test/experimental-react/package.json
```diff
@@ -4,9 +4,8 @@
   "version": "0.0.0",
   "type": "module",
   "scripts": {
-    "prepare": "rm -rf node_modules/@apollo/client-react-streaming; mkdir node_modules/@apollo/client-react-streaming && cp -r ../../packages/client-react-streaming/{package.json,dist} node_modules/@apollo/client-react-streaming",
```
jerelmiller, line 7: Did this change intentionally? I saw your other comment about making sure that we don't have multiple React versions installed, but maybe you fixed that in this PR?

jerelmiller (approved):
Looks great to me! I only really had a question about the removal of some scripts which seemed important in other PRs.

integration-test/experimental-react/package.json
phryneas: This is just a moving target 🤣 

At some point, the `prepare` script was necessary for this, then I changed it to an `exec:` approach (and removed this script), which works nicely in CI, but then it turned out that rebuilding locally becomes really annoying, so there will again be a similar script in the next PR... 🙃

PR #223, opened 2024-03-04
0.9.0 (7) - Changes to Data Transport

phryneas:
As I don't want to change the shape of the DataTransportAbstraction immediately after we released it for the first time, this one would also need to go into 0.9.0.

This switches from two very separate "request started" and "request finished" events to a stream of events with a shared identifier between queries.

This has the benefit that in the future we can easily track which events are still "unhandled", e.g. for keeping the connection open for longer.  

Also, for the first time we detect & communicate errors that happened in SSR to the browser, to restart queries there immediately.

As a positive side-effect of having a "query identifier", the query document is sent only once, not twice over.  That already plays a bit into #210.
(Technically, it's sent one more time per hook, so we're down from three times to two).


phryneas:
I think this is ready for review now :)

packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx
```diff
@@ -59,7 +67,49 @@ class ApolloClientSSRImpl<TCacheShape> extends ApolloClientBase<TCacheShape> {
...
+        };
+        queryInfo.markReady = function () {
+          subscriber.next({
+            type: "complete",
+            id,
+          });
+          subscriber.complete();
+          return markReady.apply(queryInfo, arguments as any);
+        };
+      });
```
phryneas, line 102: If someone has a better idea how to tap into link data events instead of wrapping `queryInfo` functions, I'm all ears. This is the best I could come up with.

packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx
```diff
@@ -59,7 +67,49 @@ class ApolloClientSSRImpl<TCacheShape> extends ApolloClientBase<TCacheShape> {
...
+
+      const streamObservable = new Observable<
+        Exclude<QueryEvent, { type: "started" }>
+      >((subscriber) => {
+        const { markResult, markError, markReady } = queryInfo;
+        queryInfo.markResult = function (result: FetchResult<any>) {
+          subscriber.next({
+            type: "data",
+            id,
+            result: result,
```
jerelmiller, line 82:
```suggestion
            result,
```

packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx
jerelmiller: I got nothin'

jerelmiller (approved):
Had one extremely small nit but this is great! I like the refactoring into events.

PR #225, opened 2024-03-06
Use Relative-CI to detect size changes in the Next.js integration test build.

phryneas:


phryneas:
This needs to be on `main` to start doing anything, so I guess I'll just merge this.

PR #226, opened 2024-03-06
[experiment/idea] allow enabling "forced suspense" for useQuery hook in SSR

phryneas:
This came up in https://community.apollographql.com/t/streaming-without-suspense-in-apollo-nextjs-client/7259 - essentially, we could allow users to opt `useQuery` into suspending during SSR, so `useQuery` result could also be streamed from SSR instead of just rendering a "loading" indicator in SSR.

If we actually want to do this, it should probably be documented as a "you can do this while you are migrating to suspense" approach.

### DIFF
--- a/integration-test/nextjs/src/app/cc/dynamic/useQuery_forcedSuspense/page.tsx
+++ b/integration-test/nextjs/src/app/cc/dynamic/useQuery_forcedSuspense/page.tsx
@@ -1,0 +1,40 @@
+"use client";
+
+import { useQuery } from "@apollo/experimental-nextjs-app-support/ssr";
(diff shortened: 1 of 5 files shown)


integration-test/nextjs/src/app/cc/dynamic/useQuery_forcedSuspense/page.tsx
```diff
@@ -0,0 +1,40 @@
...
+      id
+      title
+    }
+  }
+`;
+
+export const dynamic = "force-dynamic";
+
+export default function Page() {
+  enableSSRWaitForUseQuery(useApolloClient());
```
phryneas, line 25: This line should usually be part of `makeClient`, but here I wanted to enable this behaviour just for that one page.

PR #240, opened 2024-03-13
0.9.0 (8) documentation updates and "branded" error messages

phryneas:
This PR contains documentation updates and "branded" error messages to ensure that error messages match the package that has been imported.


packages/client-react-streaming/src/importErrors.test.tsx
```diff
@@ -0,0 +1,101 @@
+import assert from "node:assert";
```
phryneas, line 1: We don't have a test setup over in `@apollo/experimental-nextjs-app-support`, so we're testing this only over in this package.

Long-term it might be a good idea to mirror the test over there.

packages/experimental-nextjs-app-support/api-extractor.d.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * @packageDocumentation
```
phryneas, line 2: Nothing here yet, but we have a very long README.

packages/client-react-streaming/src/importErrors.test.tsx
phryneas: But I just tried it out, just in case:
One works:
image

The other one doesn't yet - I'll get back on that.
image

phryneas: Looking better now!
image
image

apollo-bot2:
## Detected SAST Vulnerabilities

packages/client-react-streaming/src/ManualDataTransport/ManualDataTransport.tsx
```diff
@@ -121,6 +121,55 @@ const buildManualDataTransportBrowserImpl =
...
+ *
+ * Creates a "manual" Data Transport, to be used with `WrapApolloProvider`.
+ *
+ * @remarks
+ *
+ * ### Drawbacks
+ *
+ * While this Data Transport enables streaming SSR, it has some conceptual drawbacks:
+ *
+ * - It does not have a way of keeping your connection open if your application already finished, but there are still ongoing queries that might need to be awaited transported over.
```
jerelmiller, line 135:
```suggestion
 * - It does not have a way of keeping your connection open if your application already finished, but there are still ongoing queries that might need to be transported over.
```

packages/client-react-streaming/src/ManualDataTransport/ManualDataTransport.tsx
```diff
@@ -121,6 +121,55 @@ const buildManualDataTransportBrowserImpl =
...
+ *   - This can happen if a component renders `useBackgroundQuery`, but does not read the `queryRef` with `useReadQuery`
+ *   - These "cut off" queries will be restarted in the Browser once the browser's `load` event happens
```
jerelmiller, line 137:
```suggestion
 *   - These "cut off" queries will be restarted in the browser once the browser's `load` event happens
```

packages/client-react-streaming/src/ManualDataTransport/ManualDataTransport.tsx
```diff
@@ -121,6 +121,55 @@ const buildManualDataTransportBrowserImpl =
...
+ * - If the `useInsertHtml` doesn't immediately flush data to the browser, the browser might already attain "newer" data through queries triggered by user interaction.
+ *   - This delayed behaviour is the case with the Next.js `ServerInsertedHTMLContext` and in the example Vite implementation.
+ *   - In this, case, older data from the server might overwrite newer data in the browser.
+ *   - This is minimized by simulating ongoing queries in the browser once the information of a started query is transported over.
```
jerelmiller, line 141:
```suggestion
 *   - In this, case, older data from the server might overwrite newer data in the browser. This is minimized by simulating ongoing queries in the browser once the information of a started query is transported over.
```

The 2nd sentence feels more like a continuation of the third, not so much its own point, so I think combining these makes sense.

jerelmiller (approved):
Sorry this took me so long to finish up! Looking awesome!

PR #257, opened 2024-03-26
apply documentTransforms before rerunning simulated queries

### LINKED ISSUE #256: __typename field missing in suspense queries
In version 0.8.0: `__typename` in every type by default
In version 0.9.0: `__typename` missing by default, it's included only when specific type has explicit `__typename` field to resolve in operation

even with cache config option `addTypename: true`.

phryneas:
I believe this would fix #256

While the `__typename` transform normally is applied in both queries happening during SSR, and in the Browser, in the case where a query could not fully be transported from SSR to the Browser (e.g. in the case an error was thrown on the server, or because the stream closed before it was fully transported over), these would be replayed as "simulated queries", and in that case the transform was not applied.

Hint for review: hide whitespace :)


jerelmiller (approved):
Looks great!

Issue #262, opened 2024-04-02
`NextSSRInMemoryCache` is not exported by `@apollo/experimental-nextjs-app-support/ssr`


nodegin:
In `node_modules/.pnpm/@apollo+experimental-nextjs-app-support@0.9.1_@apollo+client@3.9.10_next@14.1.4_react@18.2.0/node_modules/@apollo/experimental-nextjs-app-support/dist/ssr/index.ssr.js`

```
import { useContext } from 'react';
import { buildManualDataTransport } from '@apollo/client-react-streaming/manual-transport';
export { resetManualSSRApolloSingletons as resetNextSSRApolloSingletons } from '@apollo/client-react-streaming/manual-transport';
import { WrapApolloProvider, ApolloClient } from '@apollo/client-react-streaming';
export { DebounceMultipartResponsesLink, InMemoryCache as NextSSRInMemoryCache, RemoveMultipartDirectivesLink, SSRMultipartLink } from '@apollo/client-react-streaming';
import { ServerInsertedHTMLContext } from 'next/navigation.js';
export { useBackgroundQuery, useFragment, useQuery, useReadQuery, useSuspenseQuery } from '@apollo/client/index.js';

// src/ApolloNextAppProvider.ts

// src/bundleInfo.ts
var bundle = {
  pkg: "@apollo/experimental-nextjs-app-support/ssr",
  client: "NextSSRApolloClient",
  cache: "NextSSRInMemoryCache"
};

// src/ApolloNextAppProvider.ts
var ApolloNextAppProvider = /* @__PURE__ */ WrapApolloProvider(
  buildManualDataTransport({
    useInsertHtml() {
      const insertHtml = useContext(ServerInsertedHTMLContext);
      if (!insertHtml) {
        throw new Error(
          "ApolloNextAppProvider cannot be used outside of the Next App Router!"
        );
      }
      return insertHtml;
    }
  })
);
ApolloNextAppProvider.info = bundle;
var NextSSRApolloClient = class extends ApolloClient {
  /**
   * Information about the current package and it's export names, for use in error messages.
   *
   * @internal
   */
  static info = bundle;
};

export { ApolloNextAppProvider, NextSSRApolloClient };
//# sourceMappingURL=out.js.map
//# sourceMappingURL=index.ssr.js.map

```

NextSSRInMemoryCache is not exported.

I am using version 0.9.1

nodegin:
[...]

PR #263, opened 2024-04-02
add "edge-light" exports condition

### LINKED ISSUE #261: Window is not defined error in dev logs
After recent updates with v0.9.0 and v0.9.1 the dev console logs are being flooded with this error:

Screenshot 2024-04-01 at 12 56 24 PM

As far as I can tell it isn't causing any actual issues in the application, everything builds and functions as intended. My Dev console logs are just flooded with this over and over. Can provide any extra information that may be necessary.

phryneas:
I believe this should resolve #261 and #262.

I'm a bit hesitant to just slap a "worker" (which doesn't seem to be standardized?) on here, so I'm going for the more granular [`"edge-light"`](https://runtime-keys.proposal.wintercg.org/#edge-light)


packages/client-react-streaming/package-shape.json
```diff
@@ -26,12 +28,37 @@
       "RemoveMultipartDirectivesLink",
       "SSRMultipartLink",
       "WrapApolloProvider",
-      "resetApolloSingletons"
+      "resetApolloSingletons",
+      "built_for_ssr"
+    ],
+    "edge-light,worker,browser": [
```
phryneas, line 34: This is the order of conditions that Next.js is using, so we check for that explicitly.

See 

https://github.com/vercel/next.js/blob/bbd79ac9977e03010e09ff90d933f2ec614b031f/packages/next/src/build/webpack-config.ts#L96-L103

packages/client-react-streaming/tsup.config.ts
```diff
@@ -43,6 +43,11 @@ export default defineConfig((options) => {
       entry: {
         [output]: input,
       },
+      footer(ctx) {
+        return {
+          js: ctx.format === 'esm' ? `export const built_for_${env} = true;` : `exports.built_for_${env} = true;`
```
jerelmiller, line 48: I'm not sure I'm understanding the significance of the `build_for_<env>` variable here. Could you explain that?

jerelmiller (approved):
Changes look good. I was mainly curious about what the `bundle_for_x` was for 🙂

packages/client-react-streaming/tsup.config.ts
phryneas: Sometimes the SSR and Browser bundle have the same shape, so I'm deliberately giving them different shapes by adding an additional export, so I can safely distinguish between them in the [`package-shape.json`](https://github.com/apollographql/apollo-client-nextjs/pull/263/files#diff-766f8a75e8fbcae41995751388b3269343f7729b0ec6ee34bc41926a43f8c06b) test.

PR #266, opened 2024-04-02
try to stabilize flimsy tests

phryneas:
These tests fail from time to time, e.g. in this run: https://github.com/apollographql/apollo-client-nextjs/actions/runs/8523111570/job/23344752684

Let's see if we get them more stable?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,7 +45,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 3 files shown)


phryneas:
The infuriating thing is that these tests do not fail on localhost.

phryneas:
But downgrading my local node to v20 *does* fail the unit tests... Bumping the node version to 21 for the unit tests.

phryneas:
Logged as a node bug at https://github.com/nodejs/node/issues/52325 and pinned the node version for now.

PR #269, opened 2024-04-03
Update README to fix issue #262

nodegin:
In version [0.9.0](https://github.com/apollographql/apollo-client-nextjs/releases/tag/v.0.9.0), some packages have been separated from `@apollo/experimental-nextjs-app-support/ssr`,

this PR updates the correct usage so people can follow.

### DIFF
--- a/packages/experimental-nextjs-app-support/README.md
+++ b/packages/experimental-nextjs-app-support/README.md
@@ -5,8 +5,8 @@
 # Apollo Client support for the Next.js App Router
 
 > ❗️ This package is experimental.<br/>
-> Generally it should work well, you might run into race conditions when your Client Component is still rendering in SSR, and already making overlapping queries on the browser.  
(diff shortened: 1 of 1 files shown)


phryneas:
I'm sorry, but that's not the intended way of using it.

If you are using `@apollo/experimental-nextjs-app-support`, you should also import everything from `@apollo/experimental-nextjs-app-support` - and in this case from `@apollo/experimental-nextjs-app-support/ssr`.

We will eventually move the `/rsc` and `/ssr` imports together to `@apollo/experimental-nextjs-app-support` and rename `NextSSRApolloClient` etc. to just `ApolloClient`. That's planned for version `0.11.0`.

But you'll always only be importing from  the `@apollo/experimental-nextjs-app-support` and `@apollo/client` packages.

`"@apollo/client-react-streaming"` is only interesting for you if you want to integrate Apollo Client with another framework (or a hand-written solution).

PR #271, opened 2024-04-04
0.10.0 (4) default `connectToDevTools` to `false` in RSC and SSR builds

### LINKED ISSUE #238: Set `connectToDevTools` default to `false` in RSC and SSR builds

phryneas:
closes #238

The last PR for 0.10.0.

This defaults `connectToDevTools` to `false` in RSC and SSR builds where connecting to the DevTools is impossible anyways, but might end up in a stray `setTimeout` that hangs around for 10 seconds which uses memory unneccessarily - and e.g. prevents tests from cleaning up correctly.

No tests for this since I'm not sure how correctly to test this - but I could remove a few `connectToDevTools` from our tests, which is nice. I looked at the built artefacts though, and it's picked up correctly.


jerelmiller (approved):
Short and sweet. Love it.

PR #274, opened 2024-04-05
remove superjson dependency

### LINKED ISSUE #209: Tracking issue: investigate dropping SuperJSON dependency

phryneas:
closes #209

This is a big bundle size reduction for the library itself (see below, -44.93% on what we ship to the browser with the Next.js package), but beyond that it also cuts down a lot on data transported by the library, as SuperJSON adds a lot of extra bulk in their serialization format that we just don't need.

That said, some of our users might have non-serializable values in the cache and also want that transported over, so I'm leaving an escape hatch: `stringify` and `revive` can be configured.

### DIFF
--- a/packages/client-react-streaming/src/ManualDataTransport/ManualDataTransport.tsx
+++ b/packages/client-react-streaming/src/ManualDataTransport/ManualDataTransport.tsx
@@ -16,6 +16,7 @@ export interface ManualDataTransportOptions {
   /**
    * Prepare data for injecting into the stream by converting it into a string that can be parsed as JavaScript by the browser.
    * Could e.g. be `SuperJSON.stringify` or `serialize-javascript`.
+   * The default implementation act like a JSON.stringify that preserves `undefined`, but not do much on top of that.
    */
   stringifyForStream?: (value: any) => string;
   /**
--- a/packages/client-react-streaming/src/ManualDataTransport/ManualDataTransport.tsx
+++ b/packages/client-react-streaming/src/ManualDataTransport/ManualDataTransport.tsx
@@ -5,18 +5,30 @@ import type { RehydrationCache, RehydrationContextValue } from "./types.js";
 import type { HydrationContextOptions } from "./RehydrationContext.js";
 import { buildApolloRehydrationContext } from "./RehydrationContext.js";
 import { registerDataTransport } from "./dataTransport.js";
+import { revive, stringify } from "./serialization.js";
 
-interface BuildArgs {
+export interface ManualDataTransportOptions {
   /**
    * A hook that allows for insertion into the stream.
    * Will only be called during SSR, doesn't need to actiually return something otherwise.
    */
   useInsertHtml(): (callbacks: () => React.ReactNode) => void;
+  /**
+   * Prepare data for injecting into the stream by converting it into a string that can be parsed as JavaScript by the browser.
+   * Could e.g. be `SuperJSON.stringify` or `serialize-javascript`.
+   */
+  stringifyForStream?: (value: any) => string;
+  /**
+   * If necessary, additional deserialization steps that need to be applied on top of executing the result of `stringifyForStream` in the browser.
+   * Could e.g. be `SuperJSON.deserialize`. (Not needed in the case of using `serialize-javascript`)
+   */
+  reviveFromStream?: (value: any) => any;
 }
 
 const buildManualDataTransportSSRImpl = ({
   useInsertHtml,
-}: BuildArgs): DataTransportProviderImplementation<HydrationContextOptions> =>
+  stringifyForStream = stringify,
+}: ManualDataTransportOptions): DataTransportProviderImplementation<HydrationContextOptions> =>
   function ManualDataTransportSSRImpl({
     extraScriptProps,
     children,
@@ -29,6 +41,7 @@ const buildManualDataTransportSSRImpl = ({
       rehydrationContext.current = buildApolloRehydrationContext({
         insertHtml,
         extraScriptProps,
+        stringify: stringifyForStream,
       });
     }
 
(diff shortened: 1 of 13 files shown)


packages/client-react-streaming/src/ManualDataTransport/ManualDataTransport.tsx
```diff
@@ -5,18 +5,30 @@ import type { RehydrationCache, RehydrationContextValue } from "./types.js";
...
+export interface ManualDataTransportOptions {
   /**
    * A hook that allows for insertion into the stream.
    * Will only be called during SSR, doesn't need to actiually return something otherwise.
    */
   useInsertHtml(): (callbacks: () => React.ReactNode) => void;
+  /**
+   * Prepare data for injecting into the stream by converting it into a string that can be parsed as JavaScript by the browser.
+   * Could e.g. be `SuperJSON.stringify` or `serialize-javascript`.
+   */
```
phryneas, line 19:
```suggestion
   * Could e.g. be `SuperJSON.stringify` or `serialize-javascript`.
   * The default implementation act like a JSON.stringify that preserves `undefined`, but not do much on top of that.
   */
```

packages/client-react-streaming/src/ManualDataTransport/serialization.ts
```diff
@@ -0,0 +1,21 @@
...
+  while (stringified.includes(JSON.stringify(undefinedPlaceholder))) {
+    undefinedPlaceholder = "$" + undefinedPlaceholder;
+  }
+  return JSON.stringify(value, (_, v) =>
+    v === undefined ? undefinedPlaceholder : v
+  ).replaceAll(JSON.stringify(undefinedPlaceholder), "undefined");
+}
+
+export function revive(value: any): any {
+  return value;
```
jerelmiller, line 17: I see this `revive` function replaced the `SuperJSON.deserialize` call. 

I'm curious about the default here to just return the value unchanged instead of trying to parse it. Should this do a `JSON.parse` or equivalent of some kind?

packages/client-react-streaming/src/ManualDataTransport/serialization.test.ts
```diff
@@ -0,0 +1,41 @@
...
+          c: undefined,
+          d: "$apollo.undefined$",
+        },
+        '{"a":"$$apollo.undefined$","b":2,"c":undefined,"d":"$apollo.undefined$"}',
+      ],
+      [{ a: undefined, b: 2, c: 3 }, '{"a":undefined,"b":2,"c":3}'],
+    ] as const) {
+      test(JSON.stringify(data), () => {
+        const stringified = stringify(data);
+        const result = revive(eval(`(${htmlEscapeJsonString(stringified)})`));
```
jerelmiller, line 35: I notice you're not using `eval` in the actual usage of `revive`. Should this try and test `revive` as well in the way that it is used in the other parts of the codebase? Perhaps I'm missing something, but this feels a bit like a false positive since the default `revive` just returns the value unchanged (which means `eval` is doing all the work here).

jerelmiller (commented):
I had some questions about `revive` mostly, but I like where this is going. Super cool that you can swap out implementations if need be.

packages/client-react-streaming/src/ManualDataTransport/ManualDataTransport.tsx
```diff
@@ -5,18 +5,31 @@ import type { RehydrationCache, RehydrationContextValue } from "./types.js";
...
+   * Prepare data for injecting into the stream by converting it into a string that can be parsed as JavaScript by the browser.
+   * Could e.g. be `SuperJSON.stringify` or `serialize-javascript`.
+   * The default implementation act like a JSON.stringify that preserves `undefined`, but not do much on top of that.
+   */
+  stringifyForStream?: (value: any) => string;
+  /**
+   * If necessary, additional deserialization steps that need to be applied on top of executing the result of `stringifyForStream` in the browser.
+   * Could e.g. be `SuperJSON.deserialize`. (Not needed in the case of using `serialize-javascript`)
+   */
+  reviveFromStream?: (value: any) => any;
```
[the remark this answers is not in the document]
phryneas: No, the argument is already an object. The `string` from `stringifyForStream` gets directly injected into the stream, which makes it end up in JavaScript code.

So if the output of `stringifyForStream` is `"{ \"foo\": 1 }"`, then `reviveFromStream` will be called with an object `{ foo: 1 }`.

That's also why our own `revive` implementation is a no-op: nothing needs to be done. Our `stringifyForStream` implementation stringifies the object `{ foo: 1, bar: undefined }` as `"{ \"foo\":1, \"bar\": undefined }"` - and `revive` will then be called with the `{ foo: 1, bar: undefined }` object again.

`revive` is just "adding additional life" to an already parsed/eval'ed object. 
So you could have a `stringifyForStream` implementation that converts `{ foo: 1, bar: undefined }` to `"{ \"foo\":1, \"bar\": \"$u\" }"` and `revive` would then be called with `{ foo: 1, bar: "$u" }` and would be expected to swap out the `"$u"` with `undefined`.

Something like that can be useful with e.g. `superjson` which allows you to transport over custom types, but it's not required for our implementation or something like `serialize-javascript`, which just outputs valid JavaScript: 
`{ foo: new Date() }` is just serialized as `"{\"foo\":new Date(\"2024-04-09T08:49:42.044Z\")}"`.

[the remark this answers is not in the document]
phryneas: > "revive" sorta makes it sound like something died and you're trying to bring it back to life.

So, that's kinda the point. We have a half-alive zombie object that maybe needs some last finishing touches. I thought about `rehydrate`, but that's very confusing as React is already taken the name.

SuperJSON uses `deserialize` here, which is a tad better than `parse`, but I'm still not super happy with it.

Realistically: I don't expect that a few people will ever use this.

packages/client-react-streaming/src/ManualDataTransport/serialization.test.ts
phryneas: That would probably mean writing it into a `<script>` tag in a file, starting a browser and then pointing it to that file ;)

=> Essentially, what the browser is doing is an `eval`, and I don't think we really have a better way of testing it.

packages/client-react-streaming/src/ManualDataTransport/serialization.ts
phryneas: Before, we had `SuperJSON.stringify` and `SuperJSON.deserialize` here.

Quoting their docs:

* serialize: Serializes any JavaScript value into a JSON-compatible object.
* deserialize: Deserializes the output of Superjson back into your original value.
* stringify: Serializes and then stringifies your JavaScript value.
* parse: Parses and then deserializes the JSON string returned by stringify.

=> 
* in one direction, we need to serialize and stringify
* in the other direction, parsing is done by the browser and we only need to deserialize (if there was any additionaly serialization done in the first place)

phryneas:
I tried to explain this better. I'm really sorry for the 🤯  - I had a ton of difficulties wrapping my mind around what needs to be done here and what's already implicitly happening, too, but I didn't really know how to explain it better (without half a page of text, at least).

packages/client-react-streaming/src/ManualDataTransport/ManualDataTransport.tsx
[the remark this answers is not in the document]
phryneas: I just shared this yak shave with my RKT co-maintainers and @EskiMojo14 pointed out that `JSON.parse` has a `reviver` function argument. I probably took the name from there? 🤔

jerelmiller (approved):
Discussed my confusion in person. Makes sooo much sense. Thanks so much! Really great stuff 🎉

PR #277, opened 2024-04-08
Transport minified version of query.

phryneas:
Right now, we transport the AST of a query over the network.

That's not optimal:

Normal Query:

```gql
  query dynamicProducts {
    products {
      id
      title
      foo {
        bar {
          baz {
            yoink
          }
        }
      }
    }
  }
```

is instead transported as
```json
{
  "kind": "Document",
  "definitions": [
    {
      "kind": "OperationDefinition",
      "operation": "query",
      "name": {
        "kind": "Name",
        "value": "dynamicProducts"
      },
      "variableDefinitions": [],
      "directives": [],
      "selectionSet": {
        "kind": "SelectionSet",
        "selections": [
          {
            "kind": "Field",
            "name": {
              "kind": "Name",
              "value": "products"
            },
            "arguments": [],
            "directives": [],
            "selectionSet": {
              "kind": "SelectionSet",
              "selections": [
                {
                  "kind": "Field",
                  "name": {
                    "kind": "Name",
                    "value": "id"
                  },
                  "arguments": [],
                  "directives": []
                },
                {
                  "kind": "Field",
                  "name": {
                    "kind": "Name",
                    "value": "title"
                  },
                  "arguments": [],
                  "directives": []
                },
                {
                  "kind": "Field",
                  "name": {
                    "kind": "Name",
                    "value": "foo"
                  },
                  "arguments": [],
                  "directives": [],
                  "selectionSet": {
                    "kind": "SelectionSet",
                    "selections": [
                      {
                        "kind": "Field",
                        "name": {
                          "kind": "Name",
                          "value": "bar"
                        },
                        "arguments": [],
                        "directives": [],
                        "selectionSet": {
                          "kind": "SelectionSet",
                          "selections": [
                            {
                              "kind": "Field",
                              "name": {
                                "kind": "Name",
                                "value": "baz"
                              },
                              "arguments": [],
                              "directives": [],
                              "selectionSet": {
                                "kind": "SelectionSet",
                                "selections": [
                                  {
                                    "kind": "Field",
                                    "name": {
                                      "kind": "Name",
                                      "value": "yoink"
                                    },
                                    "arguments": [],
                                    "directives": []
                                  }
                                ]
                              }
                            }
                          ]
                        }
                      }
                    ]
                  }
                }
              ]
            }
          }
        ]
      }
    }
  ],
  "loc": {
    "start": 0,
    "end": 163
  }
}
```
and instead we could go smaller and transport
```gql
query dynamicProducts{products{id title foo{bar{baz{yoink}}}}}
```

### DIFF
--- a/packages/client-react-streaming/src/DataTransportAbstraction/printMinified.tsx
+++ b/packages/client-react-streaming/src/DataTransportAbstraction/printMinified.tsx
@@ -1,0 +1,15 @@
+import type { DocumentNode } from "@apollo/client/index.js";
+import { print } from "@apollo/client/utilities/index.js";
+
+export function printMinified(query: DocumentNode): string {
+  return (
+    print(query)
+      // replace multi-spaces with single space
+      .replace(/\s{2,}/g, " ")
+      // remove spaces that are preceeded by braces
+      .replace(/(?<=[{}])\s+/g, "")
+      // remove spaces that are preceeding braces
+      .replace(/\s+(?=[{}])/g, "")
+      .trim()
+  );
+}
--- a/packages/client-react-streaming/src/DataTransportAbstraction/printMinified.tsx
+++ b/packages/client-react-streaming/src/DataTransportAbstraction/printMinified.tsx
@@ -1,15 +1,7 @@
 import type { DocumentNode } from "@apollo/client/index.js";
 import { print } from "@apollo/client/utilities/index.js";
+import { stripIgnoredCharacters } from "graphql";
 
 export function printMinified(query: DocumentNode): string {
-  return (
-    print(query)
-      // replace multi-spaces with single space
-      .replace(/\s{2,}/g, " ")
-      // remove spaces that are preceeded by braces
-      .replace(/(?<=[{}])\s+/g, "")
-      // remove spaces that are preceeding braces
-      .replace(/\s+(?=[{}])/g, "")
-      .trim()
-  );
+  return stripIgnoredCharacters(print(query));
 }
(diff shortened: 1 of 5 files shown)


packages/client-react-streaming/src/DataTransportAbstraction/printMinified.tsx
```diff
@@ -0,0 +1,15 @@
...
+export function printMinified(query: DocumentNode): string {
+  return (
+    print(query)
+      // replace multi-spaces with single space
+      .replace(/\s{2,}/g, " ")
+      // remove spaces that are preceeded by braces
+      .replace(/(?<=[{}])\s+/g, "")
+      // remove spaces that are preceeding braces
+      .replace(/\s+(?=[{}])/g, "")
+      .trim()
```
jerelmiller, line 13: It looks like `graphql` has a [`stripIgnoredCharacters`](https://github.com/graphql/graphql-js/pull/1802) utility that handles this. Can we use that function to handle stripping the whitespace? It looks like that utility is token-aware so it should be a bit more accurate.

jerelmiller (changes requested):
My only request would be to use `stripIgnoredCharacters`, but otherwise this looks great!

packages/client-react-streaming/src/DataTransportAbstraction/printMinified.tsx
phryneas: Hmm, that pulls in about 10kb more bundle size - but since we're in SSR world here, that should be okay.
We might at some point get to the point that we pull in too much for edge functions, though (they have a limit of 1MB I believe?), so I'll make a note that that's a great place for a future size optimization.

phryneas: Ah forget about it, most of that is included in `gql` already, so it's likely fine anyways.

PR #292, opened 2024-04-22
Require sha in snapshot releases for forks

jerelmiller:
Copying the work from https://github.com/apollographql/apollo-client/pull/11794 to here as well.

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -14,7 +14,7 @@ jobs:
     if: |
       github.repository == 'apollographql/apollo-client-nextjs' &&
       (
-        github.event_name == 'workflow_dispatch' || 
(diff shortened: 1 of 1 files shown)


.github/workflows/snapshot-release.yml
```diff
@@ -33,14 +33,45 @@ jobs:
         if: github.event_name == 'issue_comment'
         id: comment-branch
 
+      - name: Get sha
+        id: parse-sha
```
alessbell, line 37: Since this workflow can be dispatched by either `workflow_dispatch` or `issue_comment` events, we can skip this step if it wasn't triggered by a comment:

```suggestion
        if: github.event_name == 'issue_comment'
        id: parse-sha
```

alessbell (approved):

phryneas:
Thank you!

PR #293, opened 2024-04-23
Update snapshot-release.yml

alessbell:
Released a new version that uses Node 20, eliminates GHA warning about deprecated use of Node 16.


phryneas (approved):
Not a lot to review here 😅

PR #294, opened 2024-04-25
add stream-utils sub-import

phryneas:
While trying to write docs I noticed that it's a big distractor to also have to explain these functions & have users copy-paste them from our docs into their codebases. Adding them as an extra entrypoint instead.


jerelmiller (approved):
Seems reasonable to me!

PR #300, opened 2024-05-15
add shared `@apollo/experimental-nextjs-app-support` entry point

phryneas:
This unifies `@apollo/experimental-nextjs-app-support/rsc` and `@apollo/experimental-nextjs-app-support/ssr` into `@apollo/experimental-nextjs-app-support`.

Once we rename the package to remove the `experimental` label, this will be the only entry point.

Also, some imports are changed:
* the `useQuery`/`useSuspenseQuery` etc. hooks have been dropped. They should be imported from `@apollo/client` instead.
* `NextSSRApolloClient` has been renamed to `ApolloClient` and is available in different variations from all environments, including RSC.
* `NextSSRInMemoryCache` has been renamed to `InMemoryCache` and is available in different variations from all environments, including RSC.
* `resetNextSSRApolloSingletons` has been renamed to `resetApolloClientSingletons`


packages/experimental-nextjs-app-support/README.md
```diff
@@ -105,21 +109,11 @@ function makeClient() {
...
-        ? ApolloLink.from([
-            // in a SSR environment, if you use multipart features like
-            // @defer, you need to decide how to handle these.
-            // This strips all interfaces with a `@defer` directive from your queries.
-            new SSRMultipartLink({
-              stripDefer: true,
-            }),
-            httpLink,
-          ])
-        : httpLink,
```
phryneas, line 122: I believe this is something that's copied blindly without any real need, so I'll omit the complexity here for now.

packages/experimental-nextjs-app-support/README.md
```diff
@@ -167,9 +161,9 @@ This package uses some singleton instances on the Browser side - if you are writ
 For that, you can use the `resetNextSSRApolloSingletons` helper:
```
jerelmiller, line 161:
```suggestion
For that, you can use the `resetApolloClientSingletons ` helper:
```

packages/experimental-nextjs-app-support/src/ssr/index.ts
```diff
@@ -1,35 +1,17 @@
-export { ApolloNextAppProvider } from "../ApolloNextAppProvider.js";
-export { resetManualSSRApolloSingletons as resetNextSSRApolloSingletons } from "@apollo/client-react-streaming/manual-transport";
-import { ApolloClient } from "@apollo/client-react-streaming";
-import { bundle } from "../bundleInfo.js";
 export {
   InMemoryCache as NextSSRInMemoryCache,
+  ApolloClient as NextSSRApolloClient,
```
jerelmiller, line 3: Do we need a way to mark these as deprecated? Or is there another way we can hint that these exports will be removed in a future version?

jerelmiller (approved):
This looks great to me! I absolutely love the simplification here.

packages/experimental-nextjs-app-support/src/ssr/index.ts
phryneas: I was thinking about that, too.... I think I'll merge this as-is and add another PR for that.

PR #301, opened 2024-05-22
Add deprecation messages to moved exports.

phryneas:


packages/experimental-nextjs-app-support/src/ssr/index.ts
```diff
@@ -1,17 +1,138 @@
...
+ * import type { TransportedQueryRef } from "@apollo/experimental-nextjs-app-support";
+ * ```
+ */
+export type TransportedQueryRef<
+  TData = unknown,
+  TVariables = unknown,
+> = _TransportedQueryRef<TData, TVariables>;
+/**
+ * @deprecated
+ * Importing this hook from this package is not necessary anymore - you can directly import it from `@apollo/client`.
```
jerelmiller, line 96:
```suggestion
 * Importing `useBackgroundQuery` from this package is no longer necessary. Import it directly from `@apollo/client` instead.
```

[nit] I'd break this up into two sentences as I think it reads a bit better. Feel free to keep what you have though if you don't like this change, especially since this is just what you see in the editor. If you do like this, the other hook deprecations could also need this change to keep this consistent.

jerelmiller (approved):
Looks great to me! This will be a welcome change 🎉

packages/experimental-nextjs-app-support/src/ssr/index.ts
phryneas: Thank you!

PR #303, opened 2024-05-23
update README to include PreloadQuery

phryneas:


packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+</PreloadQuery>
+```
+
+> The `Suspense` boundary here is optional.
+
+This example will fetch your query in RSC, and then transport the data into your Client Component cache.
+Before the child `ClientComponent` in the example renders, a "simulated network request" for this query is started in your Client Components.
+That way, if you repeat the query in your Client Component using `useSuspenseQuery` (or even `useQuery`!), it will wait for the network request in your Server Component to finish instead of making it's own network request.
+
+> Keep in mind that we recommend not to mix "client data" and "RSC data". Data fetched this way should be considerd "client data" and never be referenced in your Server Components. In fact, `PreloadQuery` will create a separate `ApolloClient` instance from the instance normally used in RSC, to prevent mixing data.
```
jerelmiller, line 188:
```suggestion
> Keep in mind that we don't recommend mixing data between Client Components and Server Components. Data fetched this way should be considered client data and never be referenced in your Server Components. `PreloadQuery` prevents mixing data by creating a separate `ApolloClient` instance using the configuration from RSC to prevent mixing data.
```

Had a few suggestions for this. Feel free to mix and match what I have with what you have if you don't entirely like this.

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+    foo: 1,
+  }}
+>
+  <Suspense fallback={<>loading</>}>
+    <ClientChild />
+  </Suspense>
+</PreloadQuery>
+```
+
+> The `Suspense` boundary here is optional.
```
jerelmiller, line 182: I wonder if we should explain why the `Suspense` boundary here is optional so we can let users know when it should be used. Is this only needed if `ClientChild` uses a `useSuspenseQuery`?

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
 
 If you want to make the most of the streaming SSR features offered by React & the Next.js App Router, consider using the [`useSuspenseQuery`](https://www.apollographql.com/docs/react/api/react/hooks-experimental/#using-usesuspensequery_experimental) and [`useFragment`](https://www.apollographql.com/docs/react/api/react/hooks-experimental/#using-usefragment_experimental) hooks.
 
+### Preloading data in RSC for usage in Client Components
+
+You can also preload data in RSC to populate the cache of your Client Components.
```
jerelmiller, line 163:
```suggestion
Starting with version 0.11, you can preload data in RSC to populate the cache of your Client Components.
```

Should we mention the version here? Totally fine if we don't add it. At the very least, I'd consider dropping the "also" here since I don't think its needed :)

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+
+For that, follow the setup steps for both RSC and Client Components as laid out in the last two paragraphs.
```
jerelmiller, line 165:
```suggestion
For that, follow the setup steps for both RSC and Client Components as laid out in the last two sections.
```

Technically the first sentence in this section is a paragraph and I don't think you're meaning to reference that one, so I'd use the word "sections" here instead to avoid confusion.

jerelmiller, line 165:
```suggestion
As a prerequisite, follow the steps outlined in the ["In RSC"](#in-rsc) and ["In Client Components and streaming SSR"](#in-client-components-and-streaming-ssr) sections to configure your client.
```

Alternatively, you can reference the section names with links to really make sure we know what to follow here :)

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+
+Then you can use the `PreloadQuery` component in your React Server Components:
```
jerelmiller, line 167:
```suggestion
For that, follow the setup steps for both RSC and Client Components as laid out in the last two paragraphs. Then you can use the `PreloadQuery` component in your React Server Components:
```

[nit] I'd combine these two paragraphs since they are part of the same thought.

jerelmiller, line 167:
```suggestion
For that, follow the setup steps for both RSC and Client Components as laid out in the last two paragraphs.

Then you can use the `PreloadQuery` component in your Server Components:
```

The [React docs](https://react.dev/reference/rsc/server-components) tend to use "Server Components" without the word "React" when referring to them so perhaps we should do the same?

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+>
+  <Suspense fallback={<>loading</>}>
+    <ClientChild />
+  </Suspense>
+</PreloadQuery>
+```
+
+> The `Suspense` boundary here is optional.
+
+This example will fetch your query in RSC, and then transport the data into your Client Component cache.
```
jerelmiller, line 184:
```suggestion
This example will fetch a query in a Server Component then transport the data into the Client Component cache.
```

Should we use "Server Component" instead of "RSC" when referring to them, similar to the [React docs](https://react.dev/reference/rsc/server-components)?

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+Before the child `ClientComponent` in the example renders, a "simulated network request" for this query is started in your Client Components.
```
jerelmiller, line 185:
```suggestion
Before the child `ClientChild` in the example renders, a "simulated network request" for this query is started in your Client Components.
```

Did you mean this component? I don't see one called `ClientComponent` in an example anywhere so just making sure this is a typo and I didn't miss something obvious.

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+
+```jsx
+<PreloadQuery
+  query={QUERY}
+  variables={{
+    foo: 1,
+  }}
+>
+  <Suspense fallback={<>loading</>}>
+    <ClientChild />
```
jerelmiller, line 177: I think it might be a good idea to show what `ClientChild` looks like here. Its not obvious how `PreloadQuery` relates to what is used in `ClientChild` without it. I assume that `ClientChild` would probably contain a `useSuspenseQuery`, but I can only guess that because I've followed this work for some time. Since you detail it in the paragraph below, I'd show the implementation of it here to really hammer home how `PreloadQuery` and `useSuspenseQuery` work together :)

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+  </Suspense>
+</PreloadQuery>
+```
+
+> The `Suspense` boundary here is optional.
+
+This example will fetch your query in RSC, and then transport the data into your Client Component cache.
+Before the child `ClientComponent` in the example renders, a "simulated network request" for this query is started in your Client Components.
+That way, if you repeat the query in your Client Component using `useSuspenseQuery` (or even `useQuery`!), it will wait for the network request in your Server Component to finish instead of making it's own network request.
```
jerelmiller, line 186: Part of me thinks the "simulated network request" is more of an implementation detail, and I'm not sure if we should care to mention that or not. I do like that you detail how this works in tandom with `useSuspenseQuery` since that is what the user cares about. Perhaps we could remove the implementation detail part of this statement and just keep to the high level overview?

```suggestion
Because we are using a `PreloadQuery` component wrapped around `ClientChild`, `useSuspenseQuery` will recognize this and avoid making a network request of its own. This also works for Client Components that use the `useQuery` hook as well!
```

(note the change assumes we show `ClientChild` somewhere with a `useSuspenseQuery`)

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+
+> Keep in mind that we recommend not to mix "client data" and "RSC data". Data fetched this way should be considerd "client data" and never be referenced in your Server Components. In fact, `PreloadQuery` will create a separate `ApolloClient` instance from the instance normally used in RSC, to prevent mixing data.
+
+#### Usage with `useReadQuery`.
+
+You can also use this approach in combination with `useReadQuery` on the client. For that, you can use this "render prop" approach to get a transportable `QueryRef` that you can pass down into your Client Components:
```
jerelmiller, line 192:
```suggestion
You can also use this approach in combination with `useReadQuery` in Client Components. Use the render prop approach to get a `QueryRef` that you can pass to your Client Component:
```

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+```jsx
+<PreloadQuery
+  query={QUERY}
+  variables={{
+    foo: 1,
+  }}
+>
+  {(queryRef) => (
+    <Suspense fallback={<>loading</>}>
+      <ClientChild queryRef={queryRef} />
```
jerelmiller, line 203: We should show the implementation of `ClientChild` as well to really highlight how this is used with `useReadQuery`

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+    </Suspense>
+  )}
+</PreloadQuery>
+```
+
+Inside of `ClientChild`, you could then call `useReadQuery` with the `queryRef` prop. The `Suspense` boundary in the example is optional.
```
jerelmiller, line 209: Making the change to the example code, I'd either eliminate this, or move some of this info to a comment in the code block itself.

> The `Suspense` boundary in the example is optional.

Like before, I'd recommend explaining when/why its optional. I'm actually surprised it is optional considering `useReadQuery` suspends on pending requests, so if I don't know when, I'm guessing our users won't either :)

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+
+#### Caveat
```
jerelmiller, line 211: Is this caveat meant for `PreloadQuery` in general or just with the last section? I see this uses a heading level 4 which puts it at the same level as the other two sections, which means its part of the "Preloading data in RSC for usage in Client Components" section.

If this is meant for the query ref section, I'd consider removing this heading and putting the next sentence in either a block quote or `NOTE` [alert](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)

packages/experimental-nextjs-app-support/README.md
```diff
@@ -154,6 +158,60 @@ export default function RootLayout({
...
+
+Keep in mind that this will look like a "new network request" to your Client Component, so make sure that the data you pass from your Server Components is not outdated, e.g. because of other caching layers you might be using.
```
jerelmiller, line 213: I'm actually a little confused at what you're getting at here, mostly because some of this seems a little contradictory. You mention its a "new" network request, but yet it might suffer from being "outdated". 

How would one make sure the data is not outdated and how would a user reconcile this? And is "other caching layers" referring to the Next.js `fetch` cache? Some more detail might be helpful here, otherwise you might consider removing this section to avoid confusion.

jerelmiller (commented):
Had some suggestions for you. Don't feel like you need to hold up the release for these changes though.

packages/experimental-nextjs-app-support/README.md
phryneas: Those links will be dead with the first rewording PR, so I'll choose the version without links :D

packages/experimental-nextjs-app-support/README.md
phryneas: I prefer to be clear here, as we also use "RSC" in places.

packages/experimental-nextjs-app-support/README.md
phryneas:
```suggestion
This example will fetch a query in RSC, and then transport the data into the Client Component cache.
```

I really want to hammer down the "RSC" bit - it's what the community has chosen to use most of the time and it's also easier to google.

packages/experimental-nextjs-app-support/README.md
phryneas: Yup, thanks!

packages/experimental-nextjs-app-support/README.md
phryneas:
```suggestion
> [!IMPORTANT]
> Keep in mind that we don't recommend mixing data between Client Components and Server Components. Data fetched this way should be considered client data and never be referenced in your Server Components. `PreloadQuery` prevents mixing server data and client data by creating a separate `ApolloClient` instance using the `makeClient` function passed into `registerApolloClient`.
```

packages/experimental-nextjs-app-support/README.md
phryneas: It applies to both notations.

packages/experimental-nextjs-app-support/README.md
phryneas: Example: Your client component cache has up-to-date data from a recent mutation, and you get incoming data from an RSC that is already cached for days because the Next.js fetch cache didn't update.

packages/experimental-nextjs-app-support/README.md
phryneas: After rereading that, I believe it's useful knowledge for the "Caveat" section.

packages/experimental-nextjs-app-support/README.md
phryneas: It's just entirely optional. It would otherwise suspend up to the next suspense boundary, somewhere in the component tree.

packages/experimental-nextjs-app-support/README.md
phryneas: 0db028488ac7b53825ccae4470285f2b32fe72be

packages/experimental-nextjs-app-support/README.md
phryneas: 0db028488ac7b53825ccae4470285f2b32fe72be

packages/experimental-nextjs-app-support/README.md
phryneas: I hope 2528e36dd178dc9bda003a26605deea553936e8f helps a bit

PR #304, opened 2024-05-28
Support next@15.0.0-rc.0

nphmuller:


phryneas:
Yup, we should definitely do that.

Thanks for the PR! :)

PR #309, opened 2024-06-05
Update README links to point to non-experimental hooks links

jer-k:
Hi, I noticed that the links to `useSuspenseQuery` and `useFragment` weren't linking to the proper place. It looks like those hooks got moved out of experimental and are stable now. I updated those links. Looks like my editor also cleared out some trailing whitespace, hopefully that is okay.

### DIFF
--- a/packages/experimental-nextjs-app-support/README.md
+++ b/packages/experimental-nextjs-app-support/README.md
@@ -5,8 +5,8 @@
 # Apollo Client support for the Next.js App Router
 
 > ❗️ This package is experimental.<br/>
-> Generally it should work well, you might run into race conditions when your Client Component is still rendering in SSR, and already making overlapping queries on the browser.  
-> This cannot be addressed from our side, but would need API changes in Next.js or React itself.  
+> Generally it should work well, you might run into race conditions when your Client Component is still rendering in SSR, and already making overlapping queries on the browser.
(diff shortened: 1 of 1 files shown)


jerelmiller (approved):
Great catch! Thanks so much for updating this.

packages/experimental-nextjs-app-support/README.md
```diff
@@ -5,8 +5,8 @@
 # Apollo Client support for the Next.js App Router
 
 > ❗️ This package is experimental.<br/>
-> Generally it should work well, you might run into race conditions when your Client Component is still rendering in SSR, and already making overlapping queries on the browser.
```
phryneas, line 8: This whitespace was unfortunately very important - in GitHub markdown, two spaces at a line end mark a `<br/>` tag 😬

jer-k: Wow, thats quite a choice by them. It looks like `\` also works to add line breaks, would you be okay with a PR that uses those as the more explicit option to get the docs back to the way they were?

jerelmiller: @jer-k yes! Apologies on my part for not realizing the spaces were intentional. I've got a command to remove trailing whitespace when I save in my editor, so I think the more explicit `\` would be welcome :).

Feel free to open a new PR if you'd like, otherwise I can tackle this later today.

PR #310, opened 2024-06-05
Use \ instead of double spaces to add line breaks

jer-k:
https://github.com/apollographql/apollo-client-nextjs/pull/309/files#r1628079676

As noted in the other PR, GFM actually supports double spaces as line breaks (https://github.github.com/gfm/#hard-line-breaks), who knew! 

This uses the other option of `\` to indicate a line break so editors don't automatically clean it up and cause issues in the future


jerelmiller (approved):
Thank you so much!

PR #334, opened 2024-07-19
fix docs example, some clarification

phryneas:
thanks to Alex for the feedback!

### DIFF
--- a/packages/experimental-nextjs-app-support/README.md
+++ b/packages/experimental-nextjs-app-support/README.md
@@ -177,7 +177,7 @@ For that, follow the setup steps for both RSC and Client Components as laid out 
 </PreloadQuery>
 ```
 
-And you can use `useSuspenseQuery` in your `ClientChild` component with the same QUERY:
+And you can use `useSuspenseQuery` in your `ClientChild` component with the same QUERY and variables:
 
 ```jsx
 "use client";
@@ -186,7 +186,7 @@ import { useSuspenseQuery } from "@apollo/client";
 // ...
 
 export function ClientChild() {
-  const { data } = useSuspenseQuery(QUERY);
+  const { data } = useSuspenseQuery(QUERY, { variables: { foo: 1 } });
   return <div>...</div>;
 }
 ```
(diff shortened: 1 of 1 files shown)


packages/experimental-nextjs-app-support/README.md
```diff
@@ -204,7 +204,7 @@ That way, if you repeat the query in your Client Component using `useSuspenseQue
 
 #### Usage with `useReadQuery`.
 
-You can also use this approach in combination with `useReadQuery` in Client Components. Use the render prop approach to get a `QueryRef` that you can pass to your Client Component:
+Just like using `useBackgroundQuery` and `useReadQuery` in place of `useSuspenseQuery` (see [the suspense documentation](https://www.apollographql.com/docs/react/data/suspense#avoiding-request-waterfalls)), you can also use `PreloadQuery` in combination with `useReadQuery` in Client Components. Use the render prop notation to get a `QueryRef` that you can pass to your Client Component:
```
jerelmiller, line 207:
```suggestion
Just like using `useBackgroundQuery` with `useReadQuery` in place of `useSuspenseQuery` (see [the suspense documentation](https://www.apollographql.com/docs/react/data/suspense#avoiding-request-waterfalls)), you can also use `PreloadQuery` in combination with `useReadQuery` in Client Components. Use the render prop notation to get a `QueryRef` that you can pass to your Client Component:
```

Changing this to `with` makes this sound a bit more like they work together.

jerelmiller: I think the thing I'm missing here is the "what" that makes `PreloadQuery` similar to `useBackgroundQuery`. I keep expecting to read something like "Just like using `useBackgroundQuery` with `useReadQuery` in place of `useSuspenseQuery` _to avoid request waterfalls_". Adding a "to x..." statement I think would help make it obvious what the connection between `PreloadQuery` and `useBackgroundQuery` is (obviously feel free to add what you were intending. Your point might not be about avoiding waterfalls).

jerelmiller, line 207:
```suggestion
Just like using [`useBackgroundQuery`](https://www.apollographql.com/docs/react/data/suspense#avoiding-request-waterfalls) and `useReadQuery` in place of `useSuspenseQuery`, you can also use `PreloadQuery` in combination with `useReadQuery` in Client Components. Use the render prop notation to get a `QueryRef` that you can pass to your Client Component:
```

Would it make more sense to move this link to `useBackgroundQuery`? I love the link to the documentation, but it feels a touch awkward to interrupt the sentence with the "see x" statement.

jerelmiller (approved):
Feel free to accept/reject any of my feedback. This is a great improvement!

packages/experimental-nextjs-app-support/README.md
phryneas: Since I'm not linking to the `uBQ` docs, but to the "avoiding waterfalls" section as a generally related concepts, I'd probably keep it separated.

phryneas:
```suggestion
Just like using `useBackgroundQuery` with `useReadQuery` in place of `useSuspenseQuery` [to avoid request waterfalls](https://www.apollographql.com/docs/react/data/suspense#avoiding-request-waterfalls), you can also use `PreloadQuery` in a similar fashion in combination with `useReadQuery` in Client Components. Use the render prop notation to get a `QueryRef` that you can pass to your Client Component:
```

But I believe this might work.

PR #346, opened 2024-08-19
docs: remove redundent imports

Netail:
Remove redundent imports in README example


jerelmiller (approved):
Good catch! Thanks so much!

Issue #132, opened 2023-11-10
Making sense of different layers of caching


devdumpling:
Hi there! Thanks for all your work on this (especially the RFC, blog posts, and discussions between you and Dan--immensely helpful to me). I've been doing a deep dive across this repo, the Next.js caching page, and various stack overflow posts trying to piece it all together.

I've been testing using this in my org's new frontend, and while I believe I've got it working, I'm still struggling to understand the different layers of caching, how these new Apollo Client instances fit into that, and ultimately how to structure queries. 

## The different layers

Going to do my best to explain based on my current understanding, feel free to correct.

**In a standard Next.js 14 app, there are four caching layers (per their docs):**

1. [React Request Memoization](https://nextjs.org/docs/app/building-your-application/caching#request-memoization) -- React's caching of `GET` fetch requests during the RSC render pass. Only caches for the lifetime of a server request during rendering. Can opt out with `AbortController`. 
2. [Next Data Cache](https://nextjs.org/docs/app/building-your-application/caching#data-cache) -- Next persistent data cache across incoming server requests. When people talk about the Next monkey patched fetch, this is what they're referring to. Can be opted out with `no-store` or `unstable_noStore`.
3. [Next Full Route Cache](https://nextjs.org/docs/app/building-your-application/caching) -- Next.js automatic static optimization. Caches static routes at build time. Caches the RSC Payload (HTML) for a static route. Can be used for non-interactive, immediate previews, then payload reconciles the client and hydrates the DOM. Opt out with route segment config (`revalidate = 0` or `dynamic = force-dynamic` or if opted out of Next Data Cache already).
4. [Next Client Router Cache](https://nextjs.org/docs/app/building-your-application/caching) -- In-memory client-side cache that stores RSC payload for a user-session. Works with prefetching to help improve back/fwd navigation for visited routes to eliminate full-page reloads and maintain state.

**Adding ApolloClient into the mix we now have:**

1. A request-scoped, RSC-only Apollo Client -- In-memory normalized Apollo cache that allows us to securely tie requests together during an RSC pass. Utilizes `React.cache` to share the instance for a single request. Seems to work with `next/cookies` in my limited testing. This works with the Next.js `Data Cache` as well.
2. An SSR/CSR Apollo Client -- Traditional ApolloProvider wrapper, but with varying logic depending on whether it's running in the SSR pass or in the browser. This works with the Next.js `Data Cache` as well.

## Questions

[...]

PR #358, opened 2024-09-04
Warn if `ssrMode` or `ssrForceFetchDelay` are used.

phryneas:
I've seen this a few times too often now.

### DIFF
--- a/packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx
+++ b/packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx
@@ -52,7 +52,10 @@ type SimulatedQueryInfo = {
 };
 
 interface WrappedApolloClientOptions
-  extends Omit<ApolloClientOptions<NormalizedCacheObject>, "cache"> {
+  extends Omit<
+    ApolloClientOptions<NormalizedCacheObject>,
+    "cache" | "ssrMode" | "ssrForceFetchDelay"
+  > {
   cache: InMemoryCache;
 }
 
@@ -68,6 +71,19 @@ class ApolloClientBase extends OrigApolloClient<NormalizedCacheObject> {
   [sourceSymbol]: string;
 
   constructor(options: WrappedApolloClientOptions) {
+    const warnings: string[] = [];
+    if ("ssrMode" in options) {
+      delete options.ssrMode;
+      warnings.push(
+        "The `ssrMode` option is not supported in %s. Please remove it from your %s constructor options."
+      );
+    }
+    if ("ssrForceFetchDelay" in options) {
+      delete options.ssrForceFetchDelay;
+      warnings.push(
+        "The `ssrForceFetchDelay` option is not supported in %s. Please remove it from your %s constructor options."
+      );
+    }
     super(
       process.env.REACT_ENV === "rsc" || process.env.REACT_ENV === "ssr"
         ? {
@@ -78,6 +94,10 @@ class ApolloClientBase extends OrigApolloClient<NormalizedCacheObject> {
     );
     const info = (this.constructor as typeof ApolloClientBase).info;
     this[sourceSymbol] = `${info.pkg}:ApolloClient`;
+
+    for (const warning of warnings) {
+      console.warn(warning, info.pkg, "ApolloClient");
+    }
 
     assertInstance(
       this.cache as unknown as InMemoryCache,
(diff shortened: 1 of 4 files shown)


packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx
```diff
@@ -79,6 +95,10 @@ class ApolloClientBase extends OrigApolloClient<NormalizedCacheObject> {
     const info = (this.constructor as typeof ApolloClientBase).info;
     this[sourceSymbol] = `${info.pkg}:ApolloClient`;
 
+    for (const warning of warnings) {
+      console.warn(warning, info.pkg, "ApolloClient");
+    }
```
phryneas, line 100: This is down here because we need access to `info`, which relies on `this`, which is only available after the `super` call - but we want to remove these options before that call.

jerelmiller (approved):
Good call on this one!

t-tiger:
@phryneas 
May I ask a question about this implementation?

Actually, this change is causing an issue with the application we are using. Specifically, with PreloadQuery, a query is first executed on the RSC, and then the same query is executed again on the client side. This is causing duplicate requests in all places where we are using useSuspenseQuery.

Until the PR is shipped, I had been able to prevent the client-side query execution after the RSC query by setting `ssrMode` and `ssrForceFetchDelay`. However, since these options have been removed, duplicate requests are now occurring.

I think there may be others facing the same issue. Do you know of any way to prevent duplicate requests between RSC and the client side? Or is there any point I might be overlooking?

t-tiger:
@phryneas 

I'm sorry for the delayed response. Allow me to ask some follow-up questions 🙇 

I think I could understand the design intent of the library. After reading [this comment](https://github.com/apollographql/apollo-client-nextjs/issues/132#issuecomment-1806048002), I see that @apollo/experimental-nextjs-app-support is designed to make the client transparent to the server-side and adopting `cache-first` for fetchPolicy by default.

The issue of duplicate requests can be reproduced by setting the fetchPolicy to `network-and-cache`, where the PreloadQuery triggers once when the page loads, and then useSuspenseQuery triggers it again.

However, our devellopment teaam use case differs. Here are the underlying conditions:

1. We are using both useSuspenseQuery and useQuery. Since the fetch policy for useQuery needs to be network-and-cache, the default fetchPolicy is also set to network-and-cache.
2. When browser back/forward occurs, server components are not re-rendered. But we need to ensure the data remains up-to-date, client-side has to refetch.

For point 1, it's difficult for developers to keep specifing `cache-first` for useSuspenseQuery and `network-and-cache` for useQuery in a large-scale development project. Therefore, it's better to unify the default fetchPolicy.

For point 2, when browser back/forward occurs, Next.js does not re-render page.tsx. If the fetchPolicy is set to cache-first, stale data will be displayed. However, due to the nature of our application, we need to refetch the data.

When these two cases combine, using `ssrMode` and `ssrForceFetchDelay` was optimal. Without them, it's not possible to achieve both.

I think this kind of situation can occur, but is it the case that the design intent of the library is to use `cache-first` and remain transparent to the server-side?

phryneas:
> Since the fetch policy for useQuery needs to be network-and-cache

That fetch policy is the right choice only in *very* rare cases, I would not recommend to use it as a default, especially not without combining it with a `nextFetchPolicy`. It might lead to a lot more network requests than you might expect.

> When browser back/forward occurs, server components are not re-rendered. But we need to ensure the data remains up-to-date, client-side has to refetch.

As a user, honestly I would be quite taken aback if I were to press the back button and get different information than I saw previously - but this is your call to make in the end.

> When these two cases combine, using ssrMode and ssrForceFetchDelay is desired. 

You might have gotten your desired result on accident, but you are really relying on undefined behaviour here that might change on a whim. This library is not designed to be used with these options and they are not supported.

---

What I read here is that you are running into problems using the hooks in certain situations with certain network policies - this seems like a [XY-Problem](https://xyproblem.info/): We are discussing `ssrMode` where we should actually discuss fetch policies.
We might need to adjust the hooks to enforce certain fetch policies during SSR (`useQuery` already is pinned to "cache-only" during SSR as it would otherwise make network requests that never wait for a result) - could you please open an issue on that topic so we can discuss that instead? (And please, let's do this in an issue and not in a PR that was closed months ago).

t-tiger:
Thank you for your response. Understood. It would have been better to open an issue rather than asking in the PR. My apologies for it.

PR #369, opened 2024-10-08
[⚠️ wip]  Experimental Transport of Async Iterables

alessbell:
Experimenting with a multipart response in a server query containing `@defer` via `PreloadQuery` that can be consumed in a client component, triggering multiple re-renders as the chunks are transported over the server <> client boundary.

### DIFF
--- a/examples/hack-the-supergraph-ssr/app/layout.tsx
+++ b/examples/hack-the-supergraph-ssr/app/layout.tsx
@@ -1,7 +1,71 @@
 import { cookies } from "next/headers";
-import { ClientLayout } from "./ClientLayout";
+import { registerApolloClient } from "@apollo/client-react-streaming";
+import {
+  gql,
+  ApolloClient,
+  InMemoryCache,
+  HttpLink,
+  TypedDocumentNode,
+} from "@apollo/client";
 
+import { ClientLayout } from "./ClientLayout";
 import { ApolloWrapper } from "./ApolloWrapper";
+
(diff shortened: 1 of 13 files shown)


phryneas:
image

This should explain the race condition you're seeing:

At (1), a simulated query is started through [`client.onQueryStarted`](https://github.com/apollographql/apollo-client-nextjs/blob/c9b29119ba59d5f8d0de78c5368c3518c4c04373/packages/client-react-streaming/src/SimulatePreloadedQuery.cc.ts#L48), and at (2) the first chunk is received, which calls [`client.onQueryProgress`](https://github.com/apollographql/apollo-client-nextjs/blob/c9b29119ba59d5f8d0de78c5368c3518c4c04373/packages/client-react-streaming/src/SimulatePreloadedQuery.cc.ts#L58) and then [resolves the simulated query](https://github.com/apollographql/apollo-client-nextjs/blob/c9b29119ba59d5f8d0de78c5368c3518c4c04373/packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx#L202-L205).

After that, `useSuspenseQuery` doesn't have the deferred fields in cache yet and thus starts a request which does not deduplicate as the simulated ongoing query has already resolved with the initial chunk.

We'll probably need some more sophisticated message-passing here so only "full data" or "final chunk" resolve the simulated query.

phryneas:
I believe this can be closed now :)

PR #382, opened 2024-10-24
fix: correct peerDependency range

jpulec:
When using the caret with a prerelease version, it won't consider even an incremented patch version to satisfy the requirement. We need to drop the prelease identifier here for this to work with the stable v15 releases.


phryneas:
That kinda conflicts with [npms own semver testing tool](https://semver.npmjs.com/):

image

and with the full range we have:

image


It also works with `yarn`, which we are using in this repo. 
What package manager are you using?

jpulec:
@phryneas Well. This is kind of odd. I was running `@apollo/experimental-nextjs-app-support@0.11.4` which lists `next: ^15.0.0-rc.0` as a peer dep, and yarn was indicating that the peer dep was unmet. I bumped to `0.11.5` and the warning went away. Maybe related to the bumping to `next: ^15.0.0` in the devDeps changed the resolution somehow.

I also found this section in the NPM docs, which is kinda confusing.
image

I didn't realize that a higher, non-prelease would satisfy, i.e. `15.0.1 > ^15.0.0-rc.0`. In any case, the semver calc is correct. Sorry for the confusion on my part.

PR #387, opened 2024-11-11
add (TeeTo|ReadFrom)ReadableStreamLink links

phryneas:


packages/client-react-streaming/src/ReadableStreamLink.tsx
```diff
@@ -0,0 +1,137 @@
...
+  return Object.assign(context, {
+    [readFromReadableStreamKey]: readableStream,
+  });
+}
+
+/**
+ * A link that allows the request to be cloned into a readable stream, e.g. for
+ * transport of multipart responses from RSC or a server loader to the browser.
+ */
+export const TeeToReadableStreamLink = new ApolloLink((operation, forward) => {
```
jerelmiller, line 53: [nit] Since the description calls this "cloning" the request, what about calling the link that?

```suggestion
export const CloneToReadableStreamLink = new ApolloLink((operation, forward) => {
```

Perhaps I'm just overly not fond of "Tee", but feel free to ignore this if you prefer the current name 😆

jerelmiller, line 53: Looking at #389, it looks like these are used internally and not actually set by the user. If that is the case, should we mark these links as `@internal`? (same goes for the read link)

```suggestion
/**
 * A link that allows the request to be cloned into a readable stream, e.g. for
 * transport of multipart responses from RSC or a server loader to the browser.
 *
 * @internal
 */
export const TeeToReadableStreamLink = new ApolloLink((operation, forward) => {
```

packages/client-react-streaming/src/ReadableStreamLink.tsx
```diff
@@ -0,0 +1,137 @@
...
+  });
+}
+
+/**
+ * Apply to a context that will be passed to a link chain containing `ReadFromReadableStreamLink`.
+ * @param readableStream
+ * @param context
+ * @returns
+ */
+export function readFromReadableStream<T extends Record<string, any>>(
```
jerelmiller, line 40: Same question as the link. Should these be marked as `@internal` since they seem to be used internally in #389?

packages/client-react-streaming/src/ReadableStreamLink.tsx
```diff
@@ -0,0 +1,137 @@
...
+  | { type: "completed" }
+  | { type: "error" };
+
+interface InternalContext {
+  [teeToReadableStreamKey]?: ReadableStreamDefaultController<ReadableStreamLinkEvent>;
+  [readFromReadableStreamKey]?: ReadableStream<ReadableStreamLinkEvent>;
+}
+
+const teeToReadableStreamKey = Symbol.for(
+  "apollo.tee.readableStreamController"
```
jerelmiller, line 15:
```suggestion
  "apollo.readableStreamController"
```

Is there anything else we'd consider putting on `apollo.tee.*`? If not, I'd simplify this to a single namespace (same goes for the read key)

jerelmiller (approved):
Code itself looks good. Most of my comments were in regards to naming.

packages/client-react-streaming/src/ReadableStreamLink.tsx
phryneas: There's a reason for the name in that context :)
https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/tee

phryneas: I'd say that everything in this package could be considered "internal" or as a building block for "build-your-own-framework". These links will e.g. be used outside of the ApolloClient implementations of this package in the React-Router implementation.

packages/client-react-streaming/src/ReadableStreamLink.tsx
phryneas: These are common enough that I see them colliding with something else in the future - it's so close to a web standard type that I would combine it with the method calling it: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultController

packages/client-react-streaming/src/ReadableStreamLink.tsx
```diff
@@ -0,0 +1,137 @@
...
+ * transport of multipart responses from RSC or a server loader to the browser.
+ */
+export const TeeToReadableStreamLink = new ApolloLink((operation, forward) => {
+  const context = operation.getContext() as InternalContext;
+
+  const controller = context[teeToReadableStreamKey];
+
+  if (controller) {
+    return new Observable((observer) => {
+      const subscription = forward(operation).subscribe(
```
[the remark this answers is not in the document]
phryneas: [400846a](https://github.com/apollographql/apollo-client-nextjs/pull/387/commits/400846ac065ca9f6144d4eeb1c35b612ba477d68)

packages/client-react-streaming/src/ReadableStreamLink.tsx
jerelmiller: Oh great! I was looking for that name somewhere, but only read the [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream) page and missed this. Makes sense now! Let's leave it.

PR #396, opened 2024-11-29
feat: mutate shortcut

Netail:
As there's a query shortcut, it would be nice to have a mutate shortcut as well. Used in server actions :)


phryneas:
Hi @Netail, thank you for the PR!

Before I get this merged, I'll have to do some investigation - I very roughly remember that the behaviour of `registerApolloClient` in Server Actions changed between versions of Next.js, so I'll have to double-check that everything works as intended there before we get this change in that encourages usage of `rAC` in server actions. (We never explicitly targeted them until now.)

I hope I can get to that next week, so please have a bit of patience  :)

Netail:
> Hi @Netail, thank you for the PR!
> 
> Before I get this merged, I'll have to do some investigation - I very roughly remember that the behaviour of `registerApolloClient` in Server Actions changed between versions of Next.js, so I'll have to double-check that everything works as intended there before we get this change in that encourages usage of `rAC` in server actions. (We never explicitly targeted them until now.)
> 
> I hope I can get to that next week, so please have a bit of patience :)

That's completely fine, please let me know, we current use getClient().mutate with NextJS 14.2.x :)

Netail:
> Before I get this merged, I'll have to do some investigation - I very roughly remember that the behaviour of `registerApolloClient` in Server Actions changed between versions of Next.js, so I'll have to double-check that everything works as intended there before we get this change in that encourages usage of `rAC` in server actions. (We never explicitly targeted them until now.)

Not only in server-actions btw, you could also use mutate in API routes (route.ts files)

phryneas:
> Not only in server-actions btw, you could also use mutate in API routes (route.ts files)

Oh gosh, honestly that sounds dangerous - it would be akin to mutating in `GET` REST endpoints - imagine the user has their browser set to auto-refresh the page regularly and it causes things to get mutated on the server all the time, or imagine a web crawler hits that mutation.

Netail:
> Oh gosh, honestly that sounds dangerous - it would be akin to mutating in `GET` REST endpoints - imagine the user has their browser set to auto-refresh the page regularly and it causes things to get mutated on the server all the time, or imagine a web crawler hits that mutation.

Not that easily hahaha, there should be some logic beforehand (maybe some verification), also can be set on different HTTP methods. But it was just an example of where such mutation function could be used.

Netail:
> So thank you for the patience and I'm sorry, but it will be some time until I can really get back to this.

No problem, take your time :)

Netail:
Ahh makes sense, thanks for checking!

If you would do something like;

```
'use server'

const { getClient, query } = registerApolloClient()
 
export async function serverAction() {
    const client = getClient()
    await query(...)
    await client.mutate(...)
}
```

Wouldn't you hit the same issue?

phryneas:
Yes, that would have the same issue - `query` as a shortcut is really designed for React Server Components, not for Server Actions. 
I just hope that this "mixed" style isn't very widespread.

Netail:
> Yes, that would have the same issue - `query` as a shortcut is really designed for React Server Components, not for Server Actions. I just hope that this "mixed" style isn't very widespread, but people stick to "use shortcut everywhere" vs "use `client` everywhere", in which case they intuitively won't use `query` in Server Actions.

In practice, I think people will choose for the query shortcut faster in server actions as it's less boilerplate. Maybe it's possible to add a check in the shortcut when it's used in a server action to throw an error? This way the user error can be minimized

phryneas:
Yeah, that would be a possibility. I wouldn't throw, but just emit a warning for now - this would probably be unexpected, and it's not an "error" per se.

phryneas:
I'll do a PR for that, but open it against the `next` branch - we want to release that soon anyways.

phryneas:
@Netail could you give the PR build from https://github.com/apollographql/apollo-client-nextjs/pull/457#issuecomment-2748523141 a try and report back if it warns correctly for you?

Netail:
> @Netail could you give the PR build from [#457 (comment)](https://github.com/apollographql/apollo-client-nextjs/pull/457#issuecomment-2748523141) a try and report back if it warns correctly for you?

Seems to work :)

PR #421, opened 2025-01-23
move integration-test monorepo into root monorepo

phryneas:
Let's see if this is less painful


phryneas:
This seems like a good way forward. We had the nested monorepos only because we were ending up with multiple React versions because we *needed* those for different integration tests - but we don't really need those anymore (it's a bit sad to see the experimental build go but meh I guess)

I think this works out a lot less frustrating in the long run

PR #422, opened 2025-02-03
TanStack Start: adjust to new streaming method post 1.99

phryneas:


phryneas:
This works now, but I'm waiting for a small last change on the TanStack side so we can use `hydrate` instead of the `useRouterState` workaround.

phryneas:
https://github.com/TanStack/router/commit/a4ae8ecd147c8ab4c3f0d544bdab5fd5ad32abbd has landed in 1.99.4 so this should be good to go if it successfully deploys.

packages/tanstack-start/src/ApolloProvider.tsx
```diff
@@ -33,72 +38,71 @@ const WrappedApolloProvider = WrapApolloProvider<{ router: AnyRouter }>(
...
-            );
-          },
-        });
       });
     } else {
-      // this needs to happen synchronously before the render continues
-      // so "loading" is kicked off before the respecitve component is rendered
-      consumeBackPressure();
+      if (onQueryEvent) {
+        registerLateInitializingQueue("__APOLLO_EVENTS__", onQueryEvent);
```
phryneas, line 87: Yay for reusing tools we already have :)

phryneas:
Yup, this seems to work nicely. Ready for review!

jerelmiller (approved):

PR #424, opened 2025-02-04
Use ReadableStreamLinks for normal data transport, too

phryneas:


packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.test.tsx
```diff
@@ -55,12 +55,12 @@ describe(
     };
     const FIRST_RESULT = { me: "User" };
     const EVENT_DATA: QueryEvent = {
-      type: "data",
+      type: "next",
```
phryneas, line 58: changing from the old `QueryEvent` type to the new one that is an extension of `ReadableStreamLinkEvent`

packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.test.tsx
```diff
@@ -313,6 +315,7 @@ describe(
         appendToBody`<div hidden id="S:1"><div id="user">User</div></div>`;
         $RS("S:1", "P:1");
 
+        await new Promise((resolve) => setTimeout(resolve, 10));
```
phryneas, line 318: Arriving request data (`simulateRequestData`) now takes a tick as it goes through the link chain instead of directly being written to the store. Without this delay, this test would have written the query first and then the data from the link would have arrived.

packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx
```diff
@@ -120,6 +100,19 @@ class ApolloClientBase extends OrigApolloClient<NormalizedCacheObject> {
...
+  setLink(newLink: ApolloLink) {
+    super.setLink.call(
+      this,
+      ApolloLink.from([
+        new ReadFromReadableStreamLink(),
+        new TeeToReadableStreamLink(),
+        newLink,
+      ])
+    );
   }
```
phryneas, line 116: All implementations (SSR, RSC, Browser) now use the same links - RSC *could* omit the `ReadFromReadableStreamLink`, but keeping all those separate details is very prone to subtle bugs; this is more reasonable.

packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx
```diff
@@ -135,150 +128,68 @@ class ApolloClientClientBaseImpl extends ApolloClientBase {
...
-    const { cacheKey, cacheKeyArr } = this.identifyUniqueQuery(hydratedOptions);
-    this.transportedQueryOptions.set(id, hydratedOptions);
 
-    const queryManager = getQueryManager(this);
+    const [controller, stream] = getInjectableEventStream();
 
-    if (
-      !queryManager["inFlightLinkObservables"].peekArray(cacheKeyArr)
-        ?.observable
-    ) {
```
phryneas, line 180: No more manually added `inFlightLinkObservables` - we really use a link, so this doesn't need to fiddle with `QueryManager` internals.

packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx
```diff
@@ -334,123 +247,71 @@ export function skipDataTransport<T extends Record<string, any>>(
...
-      this.watchQueryQueue.push({
-        event: {
-          type: "started",
-          options: serializeOptions(options),
-          id,
-        },
-        observable: streamObservable,
+      return super.watchQuery({
+        ...options,
+        context: teeToReadableStream(() => this.pushEventStream(options), {
```
phryneas, line 303: Moving this into a callback that only executes if the link is actually called means we will never inject a "started" event (or worse, a stream) when the query is deduplicated or for some other reason doesn't cause a network request.

packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx
```diff
@@ -334,123 +247,71 @@ export function skipDataTransport<T extends Record<string, any>>(
...
-  constructor(options: WrappedApolloClientOptions) {
-    super(options);
-    this.setLink(this.link);
-  }
-
-  setLink(newLink: ApolloLink) {
-    super.setLink.call(this, new TeeToReadableStreamLink().concat(newLink));
-  }
-}
+class ApolloClientRSCImpl extends ApolloClientBase {}
```
phryneas, line 314: Leaving these here even though they're empty - those class names might help in stack traces submitted by users to more quickly identify problems (and tbh., at some point I'll add back some env-specific logic here anyways)

packages/tanstack-start/src/ApolloProvider.tsx
```diff
@@ -50,19 +50,13 @@ const WrappedApolloProvider = WrapApolloProvider<{ router: AnyRouter }>(
...
-                      wrap: true,
-                      json: true,
-                    })})`
-                );
+                const script = `__APOLLO_EVENTS__.push(${jsesc(event, {
+                  isScriptContext: true,
+                  wrap: true,
+                  json: true,
+                })})`;
+                ssr.injectScript(() => script);
```
phryneas, line 59: This previously had a bug, where the `injectScript` callback would be called later and at that point in time the AC core had actually written additional properties to `event`, which meant that unneccessary data was transported over the wire.
We now stringify the object immediately.

jerelmiller (approved):
Cool stuff!

PR #425, opened 2025-02-04
TanStack: stream deferred useSuspenseQuery

phryneas:
This works for now but I really don't like the implementation 😬 

I'd really prefer if we got this to work with #424.


phryneas:
Closing in favor of #424 now that I've found a way forward there.

PR #433, opened 2025-02-13
Handle closing head tag split across chunks

### LINKED ISSUE #431: createInjectionTransformStream fails to detect closing head tag split across chunks
The `createInjectionTransformStream` method sometimes fails to inject the `"ApolloSSRDataTransport"` content into SSR HTML because it fails to detect a closing `</head>` tag that happens to be split across 2 chunks processed by the `TransformStream` here:

See:  https://github.com/apollographql/apollo-client-nextjs/blob/5293db1e8d39ef92f93f1e61ec79d08241089917/packages/client-react-streaming/src/stream-utils/createInjectionTransformStream.tsx#L86-L87

This needs to be fixed by remembering the last `"</head>".length` characters of the previous chunk, and considering this when calculating `content` and/or `index` to determine the insertion point for the `renderInjectedHtml()` call.

Note: this isn't a theoretical problem, we've seen this happen.

asearle-ta:
Fixes an issue where `createInjectionTransformStream` fails to inject any cache data into SSR output when the closing `</head>` happened to fall across the boundary between 2 chunks.

Fixes #431

### DIFF
--- a/.changeset/strange-boats-type.md
+++ b/.changeset/strange-boats-type.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client-react-streaming": patch
+---
+
+`createInjectionTransformStream`: fix handling if `</head>` is chopped up into multiple chunks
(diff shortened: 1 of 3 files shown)


phryneas:
I've added a tests and done some slight refactorings. From my side this looks good - @asearle-ta, could you please take a second look if you're happy with this? In that case I'd see that I get this shipped :)

asearle-ta:
Awesome!  Your changes look great 👍

phryneas (approved):

PR #440, opened 2025-02-17
Add an optimiziation to minimize reexecution of React components during hydration.

phryneas:


packages/client-react-streaming/src/DataTransportAbstraction/useTransportValue.tsx
```diff
@@ -17,19 +24,22 @@ export function useTransportValue<T>(value: T): T {
...
-    () => value,
-    () => valueRef.current
+    () => WhichResult.client,
+    () =>
+      valueRef.current === CLEAN
+        ? WhichResult.client
+        : equal(value, valueRef.current)
+          ? WhichResult.client
+          : WhichResult.server
   );
```
phryneas, line 38: Instead of flipping from `valueRef.current` to `value` here, this will flip from `1` to `0`, so a change from `value` in the future won't be seen as a change to the result of `uSES`.

Also, adding the `equal` check here means we can return the `value` here from the get-go if it's deeply equal with the transported value `valueRef.current` - which will further referentially stabilize values returned by this hook.

jerelmiller (approved):
💯

PR #445, opened 2025-03-07
Adjust imports to use the `@apollo/client/react` entrypoint for React-specific imports.

phryneas:


phryneas:
Tested locally with the pkg-pr-new build, I'll get this released.

Issue #248, opened 2024-03-22
Remix


phryneas:
First the good news: streaming SSR with the suspense hooks and Remix works.

All the setup for that is in this diff: https://github.com/phryneas/remix-apollo-stream-example/commit/bc43efe4a3f53d0672217cc33dfb20523af4d9c3

That said, supporting loaders is another beast.
I'll add my thoughts in this thread.

phryneas:
## Thoughts on Loaders

Loaders run independently of React.
  * on a SSR request, they run *before* streaming SSR happens in `handleRequest`.
    * as I understand it, they don't even need to run in the same process/host
  * on later page navigation, they run on the server, but no more SSR happens

As a result, our current "inject into the stream" approach won't work - we have no way of sharing an AC instance reliably between `handleRequest` and `loader`, and oftentimes there's not even a `handleRequest`

=> each loader should create their own instance of Apollo Client and `createQueryPreloader`.

```js
const loader = () => {
  const preload = createQueryPreloader(makeClient())

  const result1 = preload(query1)
  const result2 = preload(query2)

  return defer({
    result1,
    result2
  })
}
```

Now, right now `preload` creates a `queryRef` - that's hardly feasible in this case, since it can't be transported over the wire.

So we probably need to wrap `createQueryPreloader` (https://github.com/apollographql/apollo-client/pull/11719) to have a completely different functionality in SSR.

It could create a value like

```js
{
  query,
  variables,
  resultPromise
}
```

which would then be picked up by an (also wrapped) `useReadQuery` that would simulate an ongoing request (simulating the request start sooner would be great, but is probably not possible).

Currently, the [defer implementation](https://github.com/remix-run/react-router/blob/9e7486b89e712b765d947297f228650cdc0c488e/packages/router/utils.ts#L1322-L1353) doesn't support promises that are not on the top level yet.

We'd probably need buy-in from the Remix team here to be able to make according changes on their side - once we have fleshed out this story more.

LydiaF:
Hi @phryneas,

Thank you so much for this setup. I apologise if this is quite basic, but I cannot figure out how to populate the frontend Apollo cache with the data. Don't we need to do something like what is demonstrated here: https://www.apollographql.com/blog/how-to-use-apollo-client-with-remix ?

Thank you for any direction :)

Lydia

LydiaF:
Hi @phryneas,

[...]

PR #452, opened 2025-03-14
Support react-router context

vladinator1000:
Adds support for https://reactrouter.com/start/changelog#middleware-unstable

When @phryneas and I were discussing async loaders in https://github.com/apollographql/apollo-client-nextjs/issues/248#issuecomment-2718596510 I had a feeling that context might feel really nice to use.

### DIFF
--- a/integration-test/playwright/src/cc-dynamic.test.ts
+++ b/integration-test/playwright/src/cc-dynamic.test.ts
@@ -187,7 +187,7 @@ test.describe("CC dynamic", () => {
     );
   });
 
-  test.only("async loader", { tag: ["@react-router"] }, async ({ page }) => {
+  test("async loader", { tag: ["@react-router"] }, async ({ page }) => {
     await page.goto(`${base}/asyncLoader`, {
       waitUntil: "commit",
     });
--- a/integration-test/playwright/src/cc-dynamic.test.ts
+++ b/integration-test/playwright/src/cc-dynamic.test.ts
@@ -215,6 +215,8 @@ test.describe("CC dynamic", () => {
     expect(await page.getByText("loading...").count()).toBe(6);
     // deferred chunk came in
     await expect(page.getByText("cuteness overload")).toBeVisible();
+    await new Promise((resolve) => setTimeout(resolve, 500));
+
     expect(await page.getByText("Queried in SSR environment").count()).toBe(7);
     expect(await page.getByText("loading...").count()).toBe(0);
   });
--- a/integration-test/playwright/src/cc-dynamic.test.ts
+++ b/integration-test/playwright/src/cc-dynamic.test.ts
@@ -186,4 +186,20 @@ test.describe("CC dynamic", () => {
       }
     );
   });
+
+  test.only("async loader", { tag: ["@react-router"] }, async ({ page }) => {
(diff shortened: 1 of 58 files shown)


integration-test/playwright/src/cc-dynamic.test.ts
```diff
@@ -192,6 +192,22 @@ test.describe("CC dynamic", () => {
...
+    expect(await page.getByText("Queried in SSR environment").count()).toBe(1);
+    // deferred chunks still loading
+    expect(await page.getByText("loading...").count()).toBe(6);
+    // deferred chunk came in
+    await expect(page.getByText("cuteness overload")).toBeVisible();
+    expect(await page.getByText("Queried in SSR environment").count()).toBe(7);
+    expect(await page.getByText("loading...").count()).toBe(0);
+  });
+
+  test("middleware", { tag: ["@react-router"] }, async ({ page }) => {
```
vladinator1000, line 206: Same as the `async loader` test, just different endpoint.

packages/react-router/src/preloader.tsx
```diff
@@ -47,6 +50,24 @@ export function createApolloLoaderHandler(
   };
 }
 
+
+export const apolloContext = unstable_createContext<ApolloContext>();
+
+export function createApolloMiddleware(
```
vladinator1000, line 56: Same as `createApolloLoaderHandler`, but sets the context value instead of making the user wrap their loader with another function.

integration-test/react-router/app/routes/middleware.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const queryRef = apollo.preloadQuery(DEFERRED_QUERY, {
+    variables: { delayDeferred: 1000 },
+  });
+
+  await new Promise((resolve) => setTimeout(resolve, 300));
+
+  return {
+    queryRef,
+  };
+}
```
vladinator1000, line 25: Feels quite clean to use.

vladinator1000 (commented):
Notes for reviewers.

packages/react-router/src/preloader.tsx
```diff
@@ -47,6 +50,24 @@ export function createApolloLoaderHandler(
...
+  makeClient: (request: Request) => ApolloClient
+): unstable_MiddlewareFunction {
+  return async ({ request, context }) => {
+    const client = makeClient(request);
+    const preloader = createTransportedQueryPreloader(client);
+    const preloadQuery: typeof preloader = (...args) =>
+      replaceStreamWithPromiscade(preloader(...args));
```
phryneas, line 63: Since we're working in the browser here, and don't need to transport data from the server to the browser, it would be optimal if we could somehow use the same `client` instance that the React tree sees and use the normal `createQueryPreloader` of Apollo Client - see https://www.apollographql.com/docs/react/data/suspense#initiating-queries-outside-react

phryneas: Oh no, I mistook this as a feature only for client loaders. We might be getting some symmetry between server and client out of this in the end. Let's see.

phryneas:
Not closed intentionally, but I merged `next` into `main` and deleted `next`.
I'll be revisiting the context topic at a later date, feel free to reopen this PR :)

PR #454, opened 2025-03-19
update Apollo Client dev dependencies

phryneas:
This is an alternative to #452 based on what @vladinator1000 pointed out, as a discussion starter. Maybe we can also combine those somehow?

### DIFF
--- a/.changeset/funny-moose-press.md
+++ b/.changeset/funny-moose-press.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client-react-streaming": patch
+---
+
+Remove `_hydrated` property on transported queryRefs, use `WeakMap` instead.
--- a/.changeset/polite-windows-grin.md
+++ b/.changeset/polite-windows-grin.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client-integration-react-router": patch
+---
+
+Fix handling of asynchronous loaders.
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -23,9 +23,11 @@
   },
   "changesets": [
     "curvy-carrots-refuse",
+    "funny-moose-press",
     "lovely-ears-kick",
     "moody-gifts-breathe",
     "old-squids-deny",
+    "polite-windows-grin",
     "real-boats-guess",
     "rich-seahorses-reply",
     "silly-points-count",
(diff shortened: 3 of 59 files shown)


phryneas:
@vladinator1000 maybe this is something where you want to chime in? https://discord.com/channels/770287896669978684/940264701785423992/1351890281800142928

thelooter:
Fresh evidence that this got more pressing rather than less, in case it helps unstick things.

When this PR was opened, middleware and `getContext` were both `unstable_`. **In React Router 8 middleware is permanently enabled** (`future.v8_middleware` graduated) and `context` is always a `RouterContextProvider` — so context is now *the* idiomatic place for per-request state, not an experiment.

The gap is that `createApolloLoaderHandler` has `args.context` in scope and doesn't forward it:

```js
return () => (loader) => (args) => {
  const client = makeClient(args.request);   // <- args.context never passed
```

Demonstrated on react-router 8.3.0 with real middleware:

```tsx
export const authContext = createContext<string>();

export const middleware: Route.MiddlewareFunction[] = [
  async ({ context }, next) => {
    context.set(authContext, "token-resolved-in-middleware");
    return next();
  },
];

export const loader = apolloLoader<Route.LoaderArgs>()(({ context, preloadQuery }) => {
  console.log("[loader] context value:", context.get(authContext));
  return { queryRef: preloadQuery(ITEM_LIST_QUERY, { variables: { offset: 0 } }) };
});
```

```
[loader]     context value: token-resolved-in-middleware
[makeClient] args received: request | extra args: 0
```

So the loader sees middleware-derived state and the Apollo client that actually issues the query cannot. Cookie-based auth has a workaround — read `request.headers` inside `makeClient` — but anything middleware *derives* rather than forwards (a resolved user, a DB handle, a tenant, a per-request dataloader) is unreachable.

Since v8 pins the shape down, the API question that stalled this may be simpler than it was in March 2025: `context` is now unconditionally a `RouterContextProvider`, so a second parameter

```js
const client = makeClient(args.request, args.context);
```

would be backwards-compatible — existing `makeClient(request)` implementations just ignore the extra argument — without needing the `unstable_` naming or a `getContext` indirection. The wrinkle is RR7, where it's `unstable_getContext` and the type differs; whether that's worth branching on or whether this should simply be v8-and-up is a call for you.

Context for where this came from: I've been checking what's left before this package works properly on React Router 8 (#566 — peer range, with #567; #564/#565 — types). This is the one item on that list that's a genuine functional gap rather than metadata, so I wanted to add data here instead of opening a duplicate thread.

Happy to put up a PR if you can indicate the preferred shape — second parameter, an options object, or something else.

PR #455, opened 2025-03-19
Adjust README for Next.js package

phryneas:


packages/nextjs/README.md
```diff
@@ -201,6 +217,9 @@ That way, if you repeat the query in your Client Component using `useSuspenseQue
 > [!IMPORTANT]
 > Keep in mind that we don't recommend mixing data between Client Components and Server Components. Data fetched this way should be considered client data and never be referenced in your Server Components. `PreloadQuery` prevents mixing server data and client data by creating a separate `ApolloClient` instance using the `makeClient` function passed into `registerApolloClient`.
 
+> [!TIP]
+> If you are using `@defer`: `<PreloadQuery>` allows your deferred data to be streamed in chunk-by-chunk without any cutoff, just as if you would have made that same query in the browser.
```
jerelmiller, line 221: With this now a part of the library, do we have plans to deprecate the links that strip the `@defer` directive from queries?

(sorry totally unrelated to this change, but made me think of it)

jerelmiller (approved):
Looks good! Great to get rid of the experimental header 🙂

packages/nextjs/README.md
phryneas: I considered it, but it still makes sense with only `useSuspenseQuery`, so I wouldn't remove it yet.

PR #463, opened 2025-04-01
Add `dangerous_disableHookValueTransportation` to `ManualDataTransport`

phryneas:
We would need feedback from someone who runs this for a while with real traffic before we can consider adding this to the library.


TomMahle:
Hello @phryneas! First off, thank you so much for all of the incredible work across the Apollo ecosystem. I was responsible for migrating a large code base from pages router to app router and you and the other maintainers' work was an absolute life-saver!

I've recently come to notice how much bandwidth the hook serialization is consuming in our applications and I'm very interested in exploring ways to cut back on that, especially considering our applications aren't massively interactive and most of the state for that interactivity is outside the Apollo Cache. WIth that said, I'm also still somewhat wary of turning off the hook synchronization entirely. 

Would you be open to a more granular approach? Possibly opting types/fields in/out of hook serialization or opting specific calls to hooks out of the serialization? If one of these sounds like a good approach to you I'd be happy to put together a more concrete API proposal and potentially a PR.

TomMahle:
On a lighter-weight front, could we get this merged into the library with a name that makes it clear that it's experimental? It seems like the hardest part about getting some testing behind this would be applying these changes. We're specifically using the NextJS integration so having it surface there as an experimental feature as well would be ideal.

TomMahle:
For posterity I'm drawing a connection from here back to https://github.com/apollographql/apollo-client-integrations/issues/344 which is what originally got me looking down this path.

TomMahle:
For context: I'm running into a bit of a chicken-and-egg scenario where there is hesitation around utilizing this functionality in production unless it is merged into main and thus guaranteed to have an up-to-date version containing it. If we can get this merged in to main, even with something identifying it as experimental, I'll have a much easier time getting it used in prod and reporting back whatever specific metrics are desired 😄. Hopefully my findings above are enough to get past that hurdle.

phryneas:
@TomMahle you know what? Let's just get this in in-between. I'll await for a review from @jerelmiller, but from my side tbh. this is as ready as it can be.

TomMahle:
@phryneas Thank you very much!

PR #475, opened 2025-04-28
change preloadQuery tanstack to transport query preload function

DogPawHat:
With the typing of `context.preloadQuery` in the tanstack start intergration, It's possible to write code like this in a loader:

```ts
 const _movieQueryRef = preloadQuery(GET_MOVIES_SEARCH, {
   variables: {
     search: query,
     page,
   },
 });

const movieQueryRef = await _movieQueryRef.toPromise();

return { movieQueryRef };
```

(This is similar to how `ensureQueryData` would be used in an Tanstack Start and Query application)

However, because in start loaders are isomorphic, in server mode the `moveQueryRef` is typed as a `PreloadedQueryRef` when it's actually a `PreloadedQueryRef` on the browser and a `TransportedQueryRef` on the server. `toPromise` does not exist right now on `TransportedQueryRef` so you get a null error that was not caught by the Blessed Squiggly Line in Code/Cursor/Typescript LSP.

Solution in this PR is just to reverse it to always being  `TransportedQueryRef` always since it seems to be a mostly strict subset of `PreloadedQueryRef`.

### DIFF
--- a/.changeset/calm-snakes-mate.md
+++ b/.changeset/calm-snakes-mate.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client-integration-tanstack-start": patch
+---
+
+change `preloadQuery` type to `PreloadTransportedQueryFunction`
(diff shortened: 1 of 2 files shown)


phryneas (approved):
That looks good, thank you! :)

PR #501, opened 2025-09-01
support AC4 and TanStack Start 1.132

phryneas:
Version Packages (alpha)


.yarn/patches/seroval-plugins-npm-1.3.2-7e1f76a801.patch
```diff
@@ -0,0 +1,75 @@
+diff --git a/dist/cjs/development/web.cjs b/dist/cjs/development/web.cjs
+index 2aa712dd44944942ac180ca2e169b2d11b0083ea..0e9f6762696293c627de3abddb026e132beaa0f8 100644
+--- a/dist/cjs/development/web.cjs
```
phryneas, line 3: https://github.com/lxsmnsyc/seroval/pull/62

integration-test/playwright/playwright.config.ts
```diff
@@ -43,6 +43,8 @@ export default defineConfig({
     viewport: { width: 1280, height: 720 },
     ignoreHTTPSErrors: true,
     baseURL: process.env.BASE_URL || "http://localhost:3000",
+    trace: "on-first-retry",
```
phryneas, line 46: This will record traces of failed tests that I can download and replay locally.

packages/client-react-streaming/src/transportedQueryRef.ts
```diff
@@ -164,43 +194,69 @@ function createTransportedQueryRef<
...
+  client: ApolloClient,
+  userOptions: ApolloClient.WatchQueryOptions,
+  initialFetchOptions: ApolloClient.WatchQueryOptions
+) {
+  if (__DEV__) {
+    // this would be a bug in our code, `initialFetchOptions` should always be derived
+    // from `userOptions` and never change `nextFetchPolicy` or we would need additional
+    // logic to handle that case
+    invariant(
+      userOptions.nextFetchPolicy === initialFetchOptions.nextFetchPolicy
```
jerelmiller, line 230: Should we provide a reason here? Perhaps something like:

> "Encountered an unexpected bug in `@apollo/client-react-streaming`. Please file an issue"

I know this should never get hit, but if for some reason an end user encounters this, there will be no explanation to why it crashed.

packages/client-react-streaming/src/transportedQueryRef.ts
```diff
@@ -164,43 +194,69 @@ function createTransportedQueryRef<
...
+  });
+  // set the options for any future `reobserve` calls back to `userOptions` directly after that
+  observable.applyOptions({
+    ...optionsAfterCreation,
+    fetchPolicy:
+      observable.options.fetchPolicy !== initialFetchOptions.fetchPolicy
+        ? // if `fetchPolicy` was changed from `initialFetchOptions`, `nextFetchPolicy` has been applied and we're not going to touch it
+          observable.options.fetchPolicy
+        : // otherwise we restore the `userOptions.fetchPolicy` for future fetches
+          optionsAfterCreation.fetchPolicy,
```
jerelmiller, line 257:
```suggestion
      observable.options.fetchPolicy === initialFetchOptions.fetchPolicy
        ? // restore `userOptions.fetchPolicy` for future fetches
          optionsAfterCreation.fetchPolicy,
        : // otherwise `fetchPolicy` was changed from `initialFetchOptions`, `nextFetchPolicy` has been applied and we're not going to touch it
          observable.options.fetchPolicy,
```

Similarly, I find it easier to use the positive connotation since this makes it a bit more obvious that you're after restoring user provided options with this block of code.

packages/client-react-streaming/src/transportedQueryRef.ts
```diff
@@ -211,32 +267,23 @@ export function isTransportedQueryRef(
 }
 
 /** @public */
+/*
```
jerelmiller, line 270:
```suggestion
/**
```

Was this meant as a doc block or general comment? If general comment, mind if I ask that we use `//` on each line so that its obvious this isn't a typo?

packages/tanstack-start/src/Transport.ts
```diff
@@ -0,0 +1,153 @@
...
+    observable: Observable<Exclude<QueryEvent, { type: "started" }>>;
+  }): void => {
+    this.controller.enqueue(event);
+    this.ongoingStreams.add(event);
+    const finalize = () => {
+      this.ongoingStreams.delete(event);
+      if (this.shouldClose && this.ongoingStreams.size === 0 && !this.closed) {
+        this.controller.close();
+        this.closed = true;
+      }
```
jerelmiller, line 79: You might consider extracting a private `close` function that moves this logic in there. I see the same thing is used on lines 59-62 so this should probably live together (in case you need to make changes later, you don't end up forgetting to modify both).

```ts
private close() {
  if (this.ongoingStreams.size === 0 && !this.closed) {
    this.controller.close();
    this.closed = true;
  }
}
```

jerelmiller (approved):
Most of my comments are stylistic, but nothing else major.

packages/client-react-streaming/src/PreloadQuery.tsx
```diff
@@ -49,7 +49,9 @@ export async function PreloadQuery<
 }: PreloadQuery.Props<TData, TVariables> & {
   getClient: () => ApolloClient | Promise<ApolloClient>;
 }): Promise<React.ReactElement> {
-  const preloader = createTransportedQueryPreloader(await getClient());
+  const preloader = createTransportedQueryPreloader(await getClient(), {
+    ssrSideOptionOverrides: { fetchPolicy: "no-cache" },
```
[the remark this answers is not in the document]
phryneas: Yeah... it didn't really convey the intent. What do you think about
```ts
    /**
     * Overrides to the options that should happen only in the `preloader` call, but should not be transported/hydrated on the client.
     */
    notTransportedOptionOverrides?: Partial<
      ApolloClient.WatchQueryOptions<any, any>
    >;
```

packages/client-react-streaming/src/transportedQueryRef.ts
phryneas: Sure, why not.
613605d0fe1f1184d6aa24a16f97fa634e3ef04b

packages/client-react-streaming/src/transportedQueryRef.ts
```diff
@@ -116,33 +124,55 @@ export function createTransportedQueryPreloader(
...
+      });
+    } else {
+      const internalQueryRef = getInternalQueryRef(
+        client,
+        { query, ...options },
+        watchQueryOptions
+      );
+
+      return Object.assign(transportedQueryRef, wrapQueryRef(internalQueryRef));
+    }
```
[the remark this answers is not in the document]
phryneas: 049600a178b32e9b7b4ff1eb88750b6fea8372ac

packages/client-react-streaming/src/transportedQueryRef.ts
```diff
@@ -116,33 +124,55 @@ export function createTransportedQueryPreloader(
...
+      ),
+      ...ssrSideOptionOverrides,
+    };
+    if (
+      !(
+        watchQueryOptions.fetchPolicy == "no-cache" ||
+        watchQueryOptions.fetchPolicy == "network-only" ||
+        (prepareForReuse &&
+          watchQueryOptions.fetchPolicy == "cache-and-network")
+      )
```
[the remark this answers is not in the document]
phryneas: 049600a178b32e9b7b4ff1eb88750b6fea8372ac

packages/client-react-streaming/src/transportedQueryRef.ts
phryneas: Intentionally comment, not docblock. Changing to `//`.

packages/tanstack-start/src/Transport.ts
phryneas: 3fbf55462d33ee379a7de418939a76f77d9e5087

PR #549, opened 2026-04-09
Export PreloadQueryRef

teodorszeltins:
This exports `PreloadQueryRef` as a public API and adds it to the object returned by `registerApolloClient`.

Until now, it only existed as an internal helper used by `PreloadQuery` under the name `SimulatePreloadedQuery`.

## Why

`PreloadQuery` starts the query from where it appears in the render tree, so the request does not begin until React gets there.

This makes it harder to parallelise GraphQL queries with other async work (REST calls, auth checks, etc.) at the top of a Server Component.

`PreloadQueryRef` makes that flow a bit more flexible. You can create the `queryRef` early with `createTransportedQueryPreloader`, start the query right away, and render `<PreloadQueryRef>` later to hydrate it into the client tree.

## Usage

```tsx
export default async function Page() {
  const preloader = createTransportedQueryPreloader(await getClient());

  const queryRef = preloader(MY_QUERY, {
    variables: { id: "1" },
  });

  const otherData = await fetchSomeRestEndpoint();

  return (
    <ApolloWrapper>
      <PreloadQueryRef queryRef={queryRef}>
        <Suspense fallback={<>loading...</>}>
          <ClientChild queryRef={queryRef} otherData={otherData} />
        </Suspense>
      </PreloadQueryRef>
    </ApolloWrapper>
  );
}
```

## Release Notes

* **New Features**
  * Exported `PreloadQueryRef` as a public API, now available through `registerApolloClient`, enabling query-ref-first preloading workflows for improved data fetching patterns.

### DIFF
--- a/.changeset/tidy-taxis-jam.md
+++ b/.changeset/tidy-taxis-jam.md
@@ -1,0 +1,6 @@
+---
+"@apollo/client-react-streaming": minor
+"@apollo/client-integration-nextjs": minor
+---
+
+Export `PreloadQueryRef`, and add `PreloadQueryRef` to the object returned by `registerApolloClient` for query-ref-first preloading flows.
--- a/integration-test/nextjs/package.json
+++ b/integration-test/nextjs/package.json
@@ -12,6 +12,7 @@
   "dependencies": {
     "@apollo/client": "^4.0.0",
     "@apollo/client-integration-nextjs": "workspace:^",
+    "@apollo/client-react-streaming": "workspace:^",
     "@apollo/server": "^4.11.2",
     "@as-integrations/next": "^3.2.0",
     "@graphql-tools/schema": "^10.0.0",
--- a/integration-test/nextjs/src/app/rsc/client.ts
+++ b/integration-test/nextjs/src/app/rsc/client.ts
@@ -17,12 +17,13 @@ setLogVerbosity("debug");
 loadDevMessages();
 loadErrorMessages();
 
-export const { getClient, PreloadQuery, query } = registerApolloClient(() => {
-  return new ApolloClient({
-    cache: new InMemoryCache(),
-    link: delayLink.concat(
-      errorLink.concat(new IncrementalSchemaLink({ schema }))
-    ),
-    incrementalHandler: new Defer20220824Handler(),
+export const { getClient, PreloadQuery, PreloadQueryRef, query } =
+  registerApolloClient(() => {
+    return new ApolloClient({
+      cache: new InMemoryCache(),
+      link: delayLink.concat(
+        errorLink.concat(new IncrementalSchemaLink({ schema }))
+      ),
+      incrementalHandler: new Defer20220824Handler(),
+    });
   });
-});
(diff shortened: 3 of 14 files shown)


phryneas:
Hi @teodorszeltins, I'm sorry that I get to this only now!

I gotta be honest, I'm not really in favor of this change, as that split is hidden quite deliberately - it's very easy to get wrong, and due to the RSC boundaries it's extremely hard to debug if it goes wrong.

Couldn't your problem also be solved by splitting up your page component?

```ts
export default async function Page() {
  return (
    <ApolloWrapper>
      <PreloadQuery
        query={MY_QUERY}
        variables={{
          id: "1",
        }}
      >
        {(queryRef) => <InnerPage queryRef={queryRef} />}
      </PreloadQuery>
    </ApolloWrapper>
  );
}
async function InnerPage({ queryRef }) {
  const otherData = await fetchSomeRestEndpoint();
  return <ClientChild queryRef={queryRef} otherData={otherData} />;
}
```

teodorszeltins:
@phryneas Yeah, I had a feeling this was intentionally hidden.

For me/us it is that in a sizeable app, this stuff gets messy fast. We often need to preload more than one query at a time, sometimes even a variable number of them depending on the page, and that just doesn't nest cleanly. You would end up building deeper trees just to hand every ref to whoever needs it. We have fallback logic that kicks in for more than one reason. It can be a Redis cache we check first, and sometimes it's a latency budget, if a query is too slow or fails outright, we bail out and go a different route, rendering different data entirely. Right now all of that lives in plain, boring functions we can call from anywhere, including places that aren't part of the render tree at all, like Next's `generateMetadata` or a shared loader used by more than one route.

As you can tell, I am not a fan of the render-prop shape, not that our case is unusual. `preloader` to me seems easier to reason about and more flexible.

jerelmiller:
Hey @teodorszeltins 👋 

I'd like to poke at this a little bit:

> We often need to preload more than one query at a time, sometimes even a variable number of them depending on the page, and that just doesn't nest cleanly.

We generally view this as an antipattern since you can ask for what you need in a single query. We'd like to better understand why you have the need for more than one query (especially a variable number of them as you say) and why a single preload query won't work in your case. 

We're pushing [fragment colocation](https://www.apollographql.com/docs/react/data/fragments#colocating-fragments) more and more as the best practice these days and trying to get away from dropping queries all over the place. I'm afraid that introducing a solution that might encourage the use of multiple queries would propagate a pattern we're trying to move away from.

teodorszeltins:
Hey @jerelmiller, yeah, let me explain.

It helps to think of some of our pages as dashboard-style layouts with multiple modules (not literally dashboard), and they still need SSR for SEO.

We already use fragments heavily, they are great for many reasons, but here is our struggle with  single query.

Pages can be built at runtime, because of experiments or CMS config or user preferences, so we don't know the module set until later, thus variable number of queries.

We also render differently depending on whether it's a human, a bot, or a fallback case, and each of those wants different data (and even endpoint), not just a different view of the same query.

Data on the page doesn't all behave the same way either. Some parts are fine cached, others need to hit the network every time, so they need different fetch policies.

And we treat queries as critical or non-critical with their own latency budgets. If a critical one fails or times out, we want the whole page to fail or fall back. If a non-critical one fails, it's fine for that part to stay blank or get swapped for something else, while the rest of the page loads as normal.

Really, so this is less about wanting lots of small queries, and more about having an escape hatch for the cases where the normal way gets awkward.

PR #560, opened 2026-07-19
fix: don't let a late transported result overwrite newer cache data

ytoshiki:
## The bug

A field reverts to a stale value a second or so after a mutation, with no user
interaction. It's a race in the streaming SSR transport.

A query that ran on the server gets replayed in the browser as a
`network-only` fetch fed by the transported stream (`onQueryStarted`). That
result can land late, and being `network-only` it writes to the cache
unconditionally. So if the browser wrote newer data for those fields in the
meantime (a mutation result, an optimistic update, a `writeQuery`), the late
write clobbers it with the server's older snapshot.

## The fix

Before applying the first transported chunk, `onQueryProgress` checks whether
newer (non-transport) data was written to this query since it started.
If so, the stale result is dropped and the query reruns in the browser via the
existing "failed on server" path, so the cache ends up on fresh data.

To tell transport writes from real ones, `ReadFromReadableStreamLink` marks its
stream replays and the `InMemoryCache` subclass counts everything else. Using a
counter rather than diffing the cache on arrival keeps the happy path from
touching the cache at all, and means sibling transported queries, a restored
persisted cache, etc. don't trip it, while partial mutation writes still do.

## Tests

Six, on the existing harness: write-after (no refetch), write-before (the bug,
fails on `main`), plus sibling query / persisted cache / partial write / defer
cases. Also verified end-to-end in a Next.js app: the field reverts on `main`
and doesn't with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented late server-rendered query results from overwriting newer browser cache data.
  * Automatically refetches affected queries so the cache and displayed data converge on fresh server results.
  * Improved handling of partial updates and streamed responses without unnecessarily interrupting in-flight transfers.
* **Tests**
  * Added coverage for cache updates occurring before, during, and after streamed result delivery.

### DIFF
--- a/.changeset/late-transport-clobber.md
+++ b/.changeset/late-transport-clobber.md
@@ -1,0 +1,10 @@
+---
+"@apollo/client-react-streaming": patch
+---
+
+Prevent a late-arriving SSR-transported query result from overwriting newer
+data the browser cache received in the meantime (e.g. from a mutation result,
+an optimistic update or `writeQuery`). If newer data covering any part of the
+query was written while the transported result was still in flight, the query
+is refetched in the browser instead, so the cache converges on fresh server
+data rather than reverting to a stale snapshot.
--- a/packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.test.tsx
+++ b/packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.test.tsx
@@ -3,6 +3,7 @@ import { outsideOf } from "@internal/test-utils/runInConditions.js";
 import assert from "node:assert";
 import test, { afterEach, describe } from "node:test";
 import type {
+  ProgressEvent,
   QueryEvent,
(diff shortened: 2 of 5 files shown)


phryneas:
Is this something that you really encountered in real life, or a theoretical scenario?

Generally, accepting this behavior was a kinda deliberate decision - it is discussed as case 2 [in this RFC](https://github.com/apollographql/apollo-client-integrations/blob/pr/RFC-2/RFC.md#challenges-of-a-client-side-cache-in-streaming-ssr).

The same would happen without streaming, if you started a slow query on the client, then ran a mutation, and then the query result came in. 

Solving it for streaming only has the risk of making the client behave more consistent, not less, so I'm a bit hesitant on this one tbh.

ytoshiki:
I didn't hit this in production. I ran into it in the Playground!
I'll close this out. Thanks for the context

